PR #14, opened 2016-06-27
0.4.0 refactoring

pleerock:


src/decorator/decorators.ts
```diff
@@ -0,0 +1,1036 @@
...
+/**
+ * Checks if the string's length (in bytes) falls in a range.
+ */
+export function IsByteLength(min: number, max?: number, validationOptions?: ValidationOptions) {
+    return function (object: Object, propertyName: string) {
+        const args: ValidationMetadataArgs = {
+            type: ValidationTypes.IS_BYTE_LENGTH,
+            target: object.constructor,
+            propertyName: propertyName,
+            value1: min,
```
zakhenry: A refactoring thought here - it would be a cleaner api to define the `value*` into an array, ie 
`constraints: [min, max]` - that way the interface for `ValidationMetadataArgs` can easily handle *n* constraints. Also with the `...` spread operator, the arguments can be passed to the actual validator function as regular args.

pleerock: good idea

Issue #23, opened 2016-07-12
add complete es5/es6 support


pleerock:
Right now its possible to use validator with es5/es6, but we can provide more features to make this experience smoother. List of todos:
- [ ] add support of registering validation schema for target class
- [ ] add support of registering custom validations without using decorators
- [ ] make schema validation better
- [ ] cover schema validation with tests
- [ ] add samples with javascript, or javascript + babel + decorators?
- [ ] add docs how to use class-validator with javascript
- [ ] add docs how to use class-validator with system-js

NoNameProvided:
Closing in favor of #178.

PR #25, opened 2016-07-18
sync validation and api for es6/es5 usage

### LINKED ISSUE #22: Synchronous validation
Is it possible with v0.4 to reproduce the validateOrThrow functionality that was present in v0.3?

We have been using this lib to validate options passed into the constructors of modules/classes where we need to be synchronous. Unfortunately we can't target ES6 yet to get async/await support due to being pinned to node v4 and it'd be a non-trivial amount of work to rewrite the all of our bootstrapping to be asynchronous.

pleerock:
fixes #22 and #23

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,8 @@ node_js:
   - 0.12
   - 0.11
 
+before_script:
+  - typings install
+
 after_success:
   - bash <(curl -s https://codecov.io/bash)
--- a/README.md
+++ b/README.md
@@ -6,24 +6,45 @@
 [![Dependency Status](https://david-dm.org/pleerock/class-validator.svg)](https://david-dm.org/pleerock/class-validator)
 [![devDependency Status](https://david-dm.org/pleerock/class-validator/dev-status.svg)](https://david-dm.org/pleerock/class-validator#info=devDependencies)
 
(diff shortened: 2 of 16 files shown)


pleerock:
@llwt I have published it under `class-validator@next`. There are some additional changes I would like to bring before I publish as latest. Try `@next` until that. If you want to contribute it would be great if you can provide test coverage for functionality you want (or any other functionality you think is important)

sanex3339:
Right now `Errors` exists in `Errors` array even if validation was success but with empty `constraints`

```
// @IsBoolean()

 ValidationError {
    target: 
     Options {
       compact: true,
       debugProtection: false,
       debugProtectionInterval: false,
       disableConsoleOutput: false,
       encodeUnicodeLiterals: true,
       reservedNames: [],
       rotateUnicodeArray: true,
       selfDefending: true,
       sourceMap: false,
       sourceMapMode: 'separate',
       unicodeArray: true,
       unicodeArrayThreshold: -1,
       wrapUnicodeArrayCalls: true },
    value: true,
    property: 'wrapUnicodeArrayCalls',
    children: [],
    constraints: {} } 
```

sanex3339:
with alpha.2 still got
```
 ValidationError {
    value: 'separate',
    property: 'sourceMapMode',
    children: [],
    constraints: {} },
```

PR #32, opened 2016-09-13
Fixed nested validation on arrays outputting incorrectly

edcarroll:
There is a bug in nested validation - when doing nested validation on an array, the validation errors are added directly to the children of the property itself. This means that errors in different objects all appear in the same place.

This PR changes the behaviour such that each child object is given its own error (with the propertyName being their index in the original array) and its children are kept separate from the others.

Let me know if this isn't actually the expected behaviour, or whether my implementation of the fix isn't ideal.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -1,5 +1,9 @@
 # Release notes
 
+**0.6.0**
+
+* breaking change with `@ValidateNested` on arrays: Validator now groups the validation errors by sub-object, rather than them all being grouped together. See #32 for a demonstration of the updated structure.
+
 **0.5.0**
 
 * async validations must be marked with `{ async: true }` option now.
--- a/src/validation/ValidationExecutor.ts
+++ b/src/validation/ValidationExecutor.ts
@@ -169,7 +169,28 @@ export class ValidationExecutor {
             const targetSchema = typeof metadata.target === "string" ? metadata.target as string : undefined;
 
             if (value instanceof Array) {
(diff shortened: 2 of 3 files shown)


pleerock:
can you provide an example, how error output was before, and how it looks like now?

edcarroll:
No problem, below is an example:

Setup:
```ts
class SubTest {
    @IsBoolean()
    wrong = "hello";
}

class Test {
    @ValidateNested()
    subTests = [new SubTest(), new SubTest()]
}
```
Output before:
```json
[{
    "property": "subTests",
    "value": [SubTest, SubTest],
    "children": [
        {
            "property": "wrong",
            "constraints": { "isBoolean": "value should be boolean" },
            "value: "hello"
        },
        {
            "property": "wrong",
            "constraints": { "isBoolean": "value should be boolean" },
            "value: "hello"
        }
     ]
}]
```

pleerock:
okay now I understood the changes, and I think its a good change. Do you think extra tests should be added to guard this new behaviour?

edcarroll:
Awesome! I don't think so, as I have updated the nested validation test for a nested array to check for the updated behaviour.

pleerock:
okay, can you please also update a [release notes document](https://github.com/pleerock/class-validator/blob/master/doc/release-notes.md) under new *0.6.0* version changes and describe a new breaking change?

edcarroll:
I've updated the release notes describing the change. Let me know if anything further needs to be added :smile:

PR #33, opened 2016-09-14
feat: Inheritance support for Validation decorators

christophercr:
Add "InheritValidation" decorator to support validation inheritance from the parent class. Solves: https://github.com/pleerock/class-validator/issues/31


pleerock:
thank you very much for your efforts, but actually there was inheritance support in the class-validator, but there was a bug I just found and fixed [here](https://github.com/pleerock/class-validator/commit/d8794a30bdb677f8e0dc87502def41f81ca05790). Please take a look at current implementation if its okay for you then good, if no lets continue discussion here. If current implementation is okay then from your PR would be great to have a tests for this feature.

you can try latest version by `npm i class-validator@next`

christophercr:
Hi, I've tried the latest version and indeed, the bugfix solved the issue. I overlooked that piece of code when I implemented this change, sorry for that.

So I think this PR is obsolete now.

pleerock:
I'll release a new version as soon as you and @edcarroll decide if he finished with a new features and its good to go.

What about this PR, it would be great if you can adopt your test, and we merge only a test. Thank you

christophercr:
I think the test I added in this PR can be taken as is (only removing the line: _@InheritValidation(MyClass)_), since it should be tested separately to have clear visibility on this feature.

So do you want me to creat another PR with this test only or you want to copy it by yourself?

pleerock:
would be nice you just update this PR, so you'll appear in the list of contributions

christophercr:
I've updated the PR as discussed. So it contains only the unit test for this feature and also a small section in the README.md so it is clear that this is supported.

pleerock:
Thank you very much

PR #36, opened 2016-09-18
feat: Accessing parent object when using @ValidateNested

edcarroll:
Another proposal: this PR adds being able to access the parent object being validated when using the `@ValidateNested` decorator.

Example usage:

```ts
class SubPost {
    @ValidateIf(o => o.parent.example == "value")
    @IsNotEmpty()
    subExample:string;
}

class Post {
    example:string; // This is accessed by the @ValidateIf of the child class.

    @ValidateNested()
    subPost:SubPost;
}
```

This also covers the case of when the `parent` property exists and shouldn't be overwritten:

```ts
class SubPost {
    @ValidateIf(o => o.differentNameForParent.example == "value")
    @IsNotEmpty()
    subExample:string;
}

class Post {
    example:string;

    @ValidateNested({ parentProperty: "differentNameForParent" })
    subPost:SubPost;
}
```

Definition:
```ts
@ValidateNested(options?:NestedValidationOptions)
interface NestedValidationOptions {
    parentProperty?:string;
}
```

Currently if you don't specify a different `parentProperty` to the default (`"parent"`) it will overwrite any existing property; perhaps it should throw an error explaining?

As per the last feature, let me know your thoughts and any changes you think would be wise, and when you're happy I can write in the docs and tests.

Thanks! :smiley:

### DIFF
--- a/src/decorator/NestedValidationOptions.ts
+++ b/src/decorator/NestedValidationOptions.ts
@@ -1,0 +1,10 @@
+/**
+ * Options used to pass to nested validation decorators.
+ */
+export interface NestedValidationOptions {
+
+    /**
+     * Specifies the property name by which the object's parent can be accessed. Defaults to 'parent'.
+     */
+    parentProperty?: string;
+}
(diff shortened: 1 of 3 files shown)


pleerock:
I think its a bad PR, too specific and parent property looks like a monkey patch. If you need to access a parent in the nested object, you need to create bidirectional relation in your objects

edcarroll:
Objectively I am inclined to agree with you, you make some good points and certainly this can be done by a relation on the child object.

samidsco:
What are suggestions if we had object with multi nested objects inside, and validation in children depends on values in root object? Bidirectional relation is not an option, since it is simple form validation just decoupled by groups.

Issue #40, opened 2016-10-15
Validate Promise API


19majkel94:
I like this library but the API could be better. I'm talking about this:
```typescript
validate(post).then(errors => { // errors is an array of validation errors
    if (errors.length > 0) {
        console.log("validation failed. errors: ", errors);
    } else {
        console.log("validation succeed");
    }
});
```
It would be better if the `validate()` reject promise on errors. To do that, [this line](/pleerock/class-validator/blob/master/src/validation/Validator.ts#L47) should be replaced with:
```typescript
const errors = executor.stripEmptyErrors(validationErrors);
return errors.length ? Promise.reject(errors) : Promise.resolve();
```
Then we simply work with `validate()` in async&await:
```typescript
try {
    await validate(post);
    console.log("validation succeed");
} catch (errors) {
    console.log("validation failed. errors: ", errors);
}
```
And if you don't want to handle the errors (like in `routing-controllers`), you can just simply do that and it will throw error on errors:
```typescript
await validate(post);
await postRepository.persist(post);
...
```
What do you think about this? It might be a breaking change, so you should publish this as a major version (to prevent auto update on npm install for older users) or we have to create a new name for the validate function with promise rejection behavior.

pleerock:
actually I had almost same method before, its called `validateAndThrow`, but then I removed it because didn't want "double-behaviour". I like this async/await syntax more then array of errors. But since its too serious breaking change its better to add extra method instead. Maybe in next major version (1.0.0) we will replace validate method with new behaviour. (maybe we need to create a separate issue where we track everything we want to publish in next major version, and once we are ready and feel confident with all new changes, and be sure that there are no other breaking changes in a near feature - we'll publish a new major version)

19majkel94:
It can be `validateAndThrow`, thanks to ES6 desctructive import it's easy to do this:
```
import { validateAndThrow as validate } from "class-validator";
```
It's easier than creating the new function with name `validate` and exporting it in subdirectory like:
```
import { validate } from "class-validator/validation/throw";
```
> actually I had almost same method before

Can you restore it from git history or should I do the PR?

pleerock:
> Can you restore it from git history or should I do the PR?

its easier to make same changes again. Also I don't like `validateAndThrow` name, maybe you have something better in mind?

Salimlou:
Benabbou Salim.

pleerock:
hi @Salimlou 😃

Salimlou:
[...]

PR #43, opened 2016-10-16
Validate API - promise rejection

19majkel94:
Implements #40.
@pleerock, you can refactor the name to the better name as a commit to this PR and then merge 😉

### DIFF
--- a/src/validation/Validator.ts
+++ b/src/validation/Validator.ts
@@ -47,6 +47,34 @@ export class Validator {
             return executor.stripEmptyErrors(validationErrors);
         });
     }
+
+    /**
+     * Performs validation of the given object based on decorators used in given object class.
+     */
+    restrictValidate(object: Object, options?: ValidatorOptions): Promise<void>;
+
+    /**
+     * Performs validation of the given object based on validation schema.
+     */
+    restrictValidate(schemaName: string, object: Object, options?: ValidatorOptions): Promise<void>;
+
+    /**
+     * Performs validation of the given object based on decorators or validation schema.
+     */
+    restrictValidate(objectOrSchemaName: Object|string, objectOrValidationOptions: Object|ValidationOptions, maybeValidatorOptions?: ValidatorOptions): Promise<void> {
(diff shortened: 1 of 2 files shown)


src/validation/Validator.ts
```diff
@@ -47,6 +47,34 @@ export class Validator {
             return executor.stripEmptyErrors(validationErrors);
         });
     }
+
+    /**
+     * Performs validation of the given object based on decorators used in given object class.
+     */
+    restrictValidate(object: Object, options?: ValidatorOptions): Promise<void>;
```
pleerock: bad name, we need to came up with better name, need to start name with `validate*`, and something like `validateOrThrow()` or better.

src/validation/Validator.ts
```diff
@@ -47,6 +47,34 @@ export class Validator {
...
+
+    /**
+     * Performs validation of the given object based on validation schema.
+     */
+    restrictValidate(schemaName: string, object: Object, options?: ValidatorOptions): Promise<void>;
+
+    /**
+     * Performs validation of the given object based on decorators or validation schema.
+     */
+    restrictValidate(objectOrSchemaName: Object|string, objectOrValidationOptions: Object|ValidationOptions, maybeValidatorOptions?: ValidatorOptions): Promise<void> {
```
pleerock: need to do it DRY. I think this should be easy as:

```typescript
async restrictValidate(....) {
     const validationErorrs = await this.validate(...);
     if (validationErorrs.length > 0)
            return Promise.reject(validationErorrs);

     return validationErorrs;
}
```

19majkel94: It's not that simple - when you have overload signatures, you can use only them, not the 3rd function `validate(objectOrSchemaName: Object|string, objectOrValidationOptions: Object|ValidationOptions, maybeValidatorOptions?: ValidatorOptions)`. I will move the implementation to the private core method.

pleerock: ah yeah I hate this problem

PR #45, opened 2016-10-17
adding military time validator

DystopianProgrammer:
I've run into a use case whereby I'd like to validate the text of an input field being in military time format. This is support by some browsers - please see the MDN docs for time with input https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

However, there is insufficient support across all browsers for this use case. 

It uses a relatively complex regex, and rather than using the matchers validator, this would be somewhat less error-prone and easier to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,7 +655,7 @@ validator.length(str, min, max); // Checks if the string's length falls in a ran
 validator.minLength(str, min); // Checks if the string's length is not less than given number.
 validator.maxLength(str, max); // Checks if the string's length is not more than given number.
 validator.matches(str, pattern, modifiers); // Checks if string matches the pattern. Either matches('foo', /foo/i) or matches('foo', 'foo', 'i').
-validator.IsMilitaryTime(str); // Checks if the string is a valid representation of military time in the format HH:MM.
+validator.isMilitaryTime(str); // Checks if the string is a valid representation of military time in the format HH:MM.
 
 // array validation methods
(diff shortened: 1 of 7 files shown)


src/validation/ValidationTypes.ts
```diff
@@ -77,6 +77,7 @@ export class ValidationTypes {
     static MIN_LENGTH = "minLength";
     static MAX_LENGTH = "maxLength";
     static MATCHES = "matches";
+    static IS_MILITARY_TIME = "isTimeValue";
```
pleerock: name of the property and its value should match

pleerock:
please also update a README, and release notes for the next version containing this change

README.md
```diff
@@ -655,6 +655,7 @@ validator.length(str, min, max); // Checks if the string's length falls in a ran
 validator.minLength(str, min); // Checks if the string's length is not less than given number.
 validator.maxLength(str, max); // Checks if the string's length is not more than given number.
 validator.matches(str, pattern, modifiers); // Checks if string matches the pattern. Either matches('foo', /foo/i) or matches('foo', 'foo', 'i').
+validator.IsMilitaryTime(str); // Checks if the string is a valid representation of military time in the format HH:MM.
```
pleerock: first is lowercase

pleerock:
released in `0.6.4`. Thank you for your contribution! 🎉

PR #52, opened 2016-11-17
README typo

jucrouzet:
The correct property name to not include target in validation errors is `validationError` (not `error`)


pleerock:
thanks!

PR #58, opened 2017-01-03
fix ValidateNested() when value is undefined

Cyri-L:
Putting `@ValidateNested()` above missing value throws an error. Now, it first checks whether the value exists. Handy to check schemas sent via http.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "class-validator",
   "private": true,
-  "version": "0.6.6",
+  "version": "0.7.0",
   "description": "Class-based validation with Typescript / ES6 / ES5 using decorators or validation schemas. Supports both node.js and browser",
   "license": "MIT",
   "readmeFilename": "README.md",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "class-validator",
+  "name": "sw-class-validator",
   "private": true,
-  "version": "0.7.0",
+  "version": "1.0.0",
   "description": "Class-based validation with Typescript / ES6 / ES5 using decorators or validation schemas. Supports both node.js and browser",
   "license": "MIT",
   "readmeFilename": "README.md",
@@ -11,10 +11,10 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/pleerock/class-validator.git"
+    "url": "https://github.com/swanest/class-validator.git"
   },
   "bugs": {
-    "url": "https://github.com/pleerock/class-validator/issues"
+    "url": "https://github.com/swanest/class-validator/issues"
   },
   "tags": [
     "validator",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
```
pleerock: you probably want to revert this file

src/decorator/decorators.ts
```diff
@@ -53,6 +53,7 @@ export function Validate(constraintClass: Function, constraintsOrValidationOptio
  */
 export function ValidateNested(validationOptions?: ValidationOptions) {
     return function (object: Object, propertyName: string) {
+        IsDefined()(object, propertyName);
```
pleerock: can you explain this change?

src/validation/ValidationExecutor.ts
```diff
@@ -184,6 +184,11 @@ export class ValidationExecutor {
     }
     
     private nestedValidations(value: any, metadatas: ValidationMetadata[], errors: ValidationError[]) {
+
+        if (value === void 0) {
```
pleerock: always wondered what `void 0` means... can you please explain?

src/decorator/decorators.ts
Cyri-L: Yes for sure, suppose calling `ValidateNested()` on an `undefined` value. Until now, it threw an error. Now it is silent in the case of undefined, but then, we should be informed about the missing property. So basically, for each `ValidateNested()` I add a default `IsDefined()`. From now, if you perform a `ValidateNested()` on an undefined value, you'll get an error "should be defined" when calling `validate()`

src/validation/ValidationExecutor.ts
Cyri-L: `void 0` means type `null` or `undefined` :)

package.json
Cyri-L: Yes, sorry for the inconvenience. I am proceeding it.

Cyri-L: Done

src/decorator/decorators.ts
pleerock: mmm maybe its better to left use to decide if this nested property should be defined or not? So, if `@ValidateNested()` and property is missing then nothing happens. If `@ValidateNested()` + `@IsDefined()` is used on property then if its missing it will throw an error. What do you think about it?

Cyri-L: I do agree with you, let's do that. Shall I revert this section ?

pleerock: yes then

Cyri-L: Done

src/validation/ValidationExecutor.ts
YouriT: It's mostly a browser based issue. For old browsers that do not support JS 1.8.5 you could do something like:
```js
console.log(undefined);
var undefined = 1;
console.log(undefined); // Would output 1
```
Read [this](http://stackoverflow.com/a/19369078/1378525) for details

PR #59, opened 2017-01-04
Patch ValidateNested()

Cyri-L:
Calling `ValidateNested()` on an undefined value doesn't throw anymore an error


pleerock:
thank you for this PR. released in `0.6.8`

PR #74, opened 2017-03-20
Added missing constraints to IsLongerThan example

nikidimi:
IsLongerThan example is not working. I think that "constraints" should be added.


felipesabino:
@nikidimi Weird that tests failed for you as you haven't touched anything besides the docs... I guess you should rebase master to your branch, I am assuming there was some test issue from where you started and it is probably fixed now as it worked fine for me

PR #75, opened 2017-03-22
Feature - Adding IsEnum validation rule

felipesabino:
Adding enum validation decorator.

Currently there was no way of validating if a value was a valid enum and I noticed developers resorting to _hacky_ solutions like using `@Min` and `@Max` decorators to make sure value was inside the enum range. 

This PR solution aims to ease the validation of enums with number values.

Example usage

```
class MyClass {
  @IsEnum(MyEnum)
  someProperty: MyEnum;
}
```

### DIFF
--- a/test/functional/validation-functions-and-decorators.spec.ts
+++ b/test/functional/validation-functions-and-decorators.spec.ts
@@ -19,6 +19,7 @@ import {
     IsDate,
     IsDivisibleBy,
     IsEmail,
+    IsEnum,
     IsFQDN,
     IsFullWidth,
     IsHalfWidth,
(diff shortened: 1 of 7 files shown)


test/functional/validation-functions-and-decorators.spec.ts
```diff
@@ -607,6 +608,55 @@ describe("IsArray", function() {
 
 });
 
+
+describe.only("IsEnum", function() {
```
pleerock: please don't forget to remove `only` next time

pleerock:
thank you

test/functional/validation-functions-and-decorators.spec.ts
felipesabino: oh gee, sorry about that :(

PR #84, opened 2017-05-29
Add IsOptional() decorator: if value missing, validators ignored

bradyholt:
This PR adds a new decorator `@IsOptional()` which checks if a value is is empty (=== '', === null) and if so, ignores all the validators on the property.

This decorator is equivalent to `@ValidateIf` with a condition checking for a missing value.  Example:

```
@ValidateIf((o)=> o.email !== "" && o.email !== null)
email: string;
```

This allows for a value to be completely optional for the purpose of validation, but if supplied, to be validated with existing validators.


pleerock:
thanks for this PR!

Issue #61, opened 2017-01-12
@Length minimum validation off by 1


jerradpatch:
Sorry, to give keep lodging issues, I am just using this validation lib a lot recently. 

so I have this,
```javascript
  @Length(2, 40)
  @IsString()
  BRIEF_DESCRIPTION : string =null;
```
when I enter '1' I get no validation error. When I have this 
```javascript
  @Length(2, 40)
  @IsString()
  BRIEF_DESCRIPTION : string =null;
```
and enter '1' I get an error 'BRIEF_DESCRIPTION must be longer than or equal to 2 characters'.

In addition, it seems @MinLength(min: number), has the same issue? Where a validation error isn't returned unless @MinLength(2) or greater?

using version 0.6.0

pleerock:
fixed by #96

PR #96, opened 2017-06-26
minLength and maxLength inclusive error messages

jerradpatch:
I actually fixed this in my local copy at one time, then updated the package and lost it. Then someone opened a jira ticket complaining to me about the same issue I complained about (this pr issue)... so I finally decided to fix it.


jerradpatch:
Updated tests as well and then ran the package on my project to confirm its working

jerradpatch:
related to issue 
#61
but I didnt fix length... just min and max

pleerock:
Thank you! Will release it in next version

PR #102, opened 2017-07-05
add support for undefined value in isOptional

### LINKED ISSUE #101: IsOptional Typescript dialect
In Typescript the *optional* word is used to mean that an property in the object can be `undefined`.

> Optional Properties
> Not all properties of an interface may be required. Some exist under certain conditions or may not be there at all. These optional properties are popular when creating patterns like “option bags” where you pass an object to a function that only has a couple of properties filled in.

[From the Typescript hand book](https://www.typescriptlang.org/docs/handbook/interfaces.html)

But the `IsOptional` decorator checks if its an empty `string` and not `null`.
```typescript
return object[propertyName] !== "" && object[propertyName] !== null;
```
[Source](https://github.com/pleerock/class-validator/blob/6aec9968eb8fc1677e82eafd8b78fdb18e803824/src/decorator/decorators.ts#L205)

In my opinion `IsOptional` should check if it is `undefined` (and maybe `null`) and the current version should be renamed.

rcodonnell:
fixes #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,7 +288,7 @@ import {ValidateIf, IsNotEmpty} from "class-validator";
 
 export class Post {
     otherProperty:string;
-    
+
     @ValidateIf(o => o.otherProperty === "value")
     @IsNotEmpty()
     example:string;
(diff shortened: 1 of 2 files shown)


rcodonnell:
@pleerock I was wondering if you would look over this PR

kavinvin:
As @adnan-kamili said, why does `@IsOptional` also treat `empty string` as a missing value while it's actually defined but only empty? It also allow many validation hole. For example, if I use @Contrains which check if an input contains the substring and then add `@Optional`, when a user enter an empty string it will be qualified as a valid input.

It would be great if it simply check the same constraint of @IsDefined as its counterpart not @IsNotEmpty
https://github.com/pleerock/class-validator/blob/master/src/validation/Validator.ts#L261

adnan-kamili:
@pleerock allowing null or empty string to bypass in `@IsOptional` is a security hole, it would allow `null` or `""` values to bypass for every datatype like boolean, Date, number. To use this decorator one needs to add additional validation checks in the code which sort of defeats the purpose.

rcodonnell:
@pleerock You merge but never released it to npm.
Anyway you can do that please.

PR #110, opened 2017-07-27
@IsOptional() and skipMissingProperties should have similar behaviour

adnan-kamili:
Allowing empty string in `@IsOptional ` opens a security hole in validation for datatypes other than string type. Additionally, it should behave similarly to "skipMissingProperties" which also skips validation only when the property is null or undefined


adnan-kamili:
@pleerock Due you think this can make into the master branch, if you agree with the behavior, if not we can handle it ourselves in the code

NoNameProvided:
Hi @adnan-kamili !

@pleerock is pretty busy with moving the TypeORM project forward right now, for what I am really grateful. I am sure he will review this PR and comment or merge it as soon as he has time for it.

In the meantime I advice you to handle it in your own codebase.

Issue #124, opened 2017-10-02
Catching Invalid date


florentsuc:
https://github.com/pleerock/class-validator/blob/83511f216bc6e413e3cdbe964556b3fc821f92b8/src/validation/Validator.ts#L322

`value instanceof Date` is true for an invalid date. We should check if the date is valid.

I'm submitting a PR.

NoNameProvided:
Closed in #125

PR #125, opened 2017-10-02
fix(isDate): checking if the date is valid (issue #124)

florentsuc:
`value instanceof Date` is true for an invalid date. We should check if the date is valid.


NoNameProvided:
Hey @florentsuc 

Sorry for the late merge! Thanks for your contribution!

closes #124

Issue #83, opened 2017-05-23
ValidateNested not working


dkirchhof:
Hello,

i have something like this:
```ts
class Day
{
    @IsMilitaryTime()
    from: string;

    @IsMilitaryTime()
    to: string;
}

export class DailyCalendar
{
    @IsDefined() // do i need it?
    @ValidateNested()
    monday: Day;
    ...
}
```

When calling the route with this body: `{ "monday": { "from": "foo" } }` or `{ "monday": {  } }` it doesn't throw an error.
Also when not using `@IsDefined()` i can call the route without `monday` property and it doesn't throw an error.

The class will be created, but not validated.

NoNameProvided:
> @IsDefined() // do i need it?

You dont, it should throw by default when using `@ValidateNested()`.

Do you use this with pleerock/routing-controllers? In previous versions, validation was turned off by default. Can you try with the latest version? (0.7.0-alpha.11)

mrgrimmig:
Hello, 
yes, I use it with routing-controllers. I updated from version 0.7.0-alpha.10 to 0.7.0-alpha.11, but didn't fix it.
Anyways I set the validation property to true in the createServer-function.

Well, the validation works in the main class.
E.g.: when I change my class to this:
```ts
...
@IsDefined()
monday: any;
...
```
and I don't provide the monday attribute in my http call, it will throw an BadRequestError.
It's just the nested validation, which doesn't work.

albertkim:
I am experiencing this issue right now as well - `ValidateNested()` doesn't seem to catch the errors on the nested objects. I have tried with both `@IsDefined()` and without it.

Version `0.7.1`

jerradpatch:
Post full code of how you substantiate the object to be validated. I use this annotation without issues. I believe you are not initializing the Monday obj properly.

19majkel94:
> I suspect that this was an issue because I was just creating an object with the right fields, not an instance of a class, which is probably needed in order for the decorators to do their thing. I now use ES6 classes properly, and @ValidateNested() appears to be working correctly.

Yes, validation works only with instances of class decorated with constraints - for validating plain JS object I recommend you to use [class-transformer-validator](https://github.com/19majkel94/class-transformer-validator) 😉

ruiming:
I have the same problem. This work for me:
```typescript
export class ICreateOrUpdateActivity implements IActivity {
  @IsString() title: string
[...]

PR #133, opened 2017-11-04
Making nested validation to work with null

franza:
Fixing #83. Added simple check on null and undefined.


franza:
Fixing #83. Added simple check on `null` and `undefined`.

MistyKuu:
When can we except it to be merged?

NoNameProvided:
Hi guys! 

This is already fixed and will be released soon. Now the `@ValidateNested` decorator correctly generates a `ValidationError` if the value is not an array or object.

PR #137, opened 2017-11-13
[fix] @ValidateNested() used to throw Error when the property being v…

aviranco:
@ValidateNested() used to throw Error when the property being validated is not an object. This causes all the other validation errors to be ignored, and returning exception with no details regarding this validation error. This fix removes the error throw, and can be used together with custom "IsObject" validator to add indicative ValidationError


NoNameProvided:
I don't think this will be merged without adding some return value. I see your point, and I agree we should raise a validation error, instead of throwing an exception, however, we should not simply just remove the exception.

I see after removing this line the tests still pass, so we have no coverage here, this should be fixed also.

PR #157, opened 2018-01-15
Fixed error in validation example

samolesen:
The validation group example in the readme, does not produce the same results as described.
I assume the comment is wrong and not the implementation.


NoNameProvided:
Thanks!

PR #161, opened 2018-01-21
feat: add `@InheritValidation` decorator to copy validation metadata

VinceOPS:
Hi,

Using [nestjs](https://nestjs.com/) recently, we've been through the frustration of duplicating a lot of validation decorators in our DTO's as we're adding more endpoints for our users, in order to do "partial updates" (and other operations requiring partial DTO's).

Explanations (using fake codes, for the sake of the example):

```typescript
export class UserDto {
  @isNumber()
  readonly id: number;
  
  @isEmail()
  readonly email: string;

  @IsString()
  @MaxLength(30)
  readonly firstName: string;

  @IsString()
  @Length(8, 32)
  readonly password: string;
}
```

Let's assume this is our complete User DTO.
At some point, we'll propose our users to update, for instance, their password. We can either create a DTO with all these fields set as Optional, and copy/paste all other validation rules. Or we can create a new DTO with only the fields we expect to be used by the client app, but we still need to copy/paste all other validation rules. And for our administrators (using the backoffice), we'll have to create different DTO's for different fields, again.
By the way, we enjoy using **specific** classes as return-types for our Controllers, and it helps a lot using Nestjs's [swagger](https://docs.nestjs.com/recipes/swagger) module.

The issue concerns the duplication of the decorators and the maintainance of all the rules.

This PR intends to fix it using a new decorator used to copy the `ValidationMetadata` of a given property.

Following our previous example, we would simply do:

```typescript
import {UserDto} from "./somewhere/user.dto.ts";

class UpdatePasswordUserDto {
  @InheritValidation(UserDto)
  readonly password: string;
}
```

But it could also be used to "share" validation metadatas between properties of the same class (let's say, "firstName" and "lastName").

I also shared my unit tests while I was at it. However, I'm using Jest (and ts-jest) in most of my projects, so this won't work with Mocha. If you like the idea and want to accept it, I'll rewrite my unit tests to work with Mocha.
(In the mean time, you can check the unit tets work (**they do** :-) ) by running jest against the file inherit-validation.spec;ts).

Thanks for all your work!

Regards

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@types/chai": "^3.4.35",
     "@types/chai-as-promised": "0.0.30",
     "@types/gulp": "^4.0.2",
+    "@types/jest": "^22.0.1",
     "@types/lodash": "^4.14.93",
     "@types/mocha": "^2.2.40",
     "@types/node": "^8.10.11",
--- a/package.json
+++ b/package.json
@@ -23,12 +23,14 @@
     "typescript-validator"
   ],
   "dependencies": {
+    "lodash": "^4.17.4",
     "validator": "9.2.0"
   },
   "devDependencies": {
     "@types/chai": "^3.4.35",
     "@types/chai-as-promised": "0.0.30",
     "@types/gulp": "^4.0.2",
+    "@types/lodash": "^4.14.93",
     "@types/mocha": "^2.2.40",
     "@types/node": "^8.10.11",
     "@types/sinon": "^1.16.36",
--- a/src/decorator/inherit-validation/inherit-validation.spec.ts
+++ b/src/decorator/inherit-validation/inherit-validation.spec.ts
@@ -1,0 +1,182 @@
+import * as _ from "lodash";
+import {getFromContainer} from "../../container";
+import {validate} from "../../index";
+import {MetadataStorage} from "../../metadata/MetadataStorage";
+import {ValidationMetadata} from "../../metadata/ValidationMetadata";
+import {IsString, MaxLength, IsNumber, Max, IsEmail, IsOptional} from "../decorators";
+import InheritValidation from "./inherit-validation";
+
+/**
+ * Used as a base for validation, in order for partial classes
+ * to pick validation metadatas, property by property.
+ */
+class Dto {
+  @IsNumber()
+  @Max(999)
+  readonly id: number;
+
+  @IsString()
+  @MaxLength(10)
+  readonly name: number;
+
+  @IsEmail()
+  @IsOptional()
+  readonly email: string;
+}
+const validationsCount = 6;
+
+describe("@InheritValidation", () => {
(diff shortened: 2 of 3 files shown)


VinceOPS:
NB: `gulp tests` runs successfully once the Jest tests are removed.

VinceOPS:
@NoNameProvided Hi! Could you give me your opinion about this decorator/idea when you get a chance? Thanks! (I can remove the Jest Tests and rewrite them for Mocha, so the Travis CI will succeed, if you agree with the feature).

NoNameProvided:
The idea is great thanks for the contribution! 🎉 

 I have one question:
```ts
const toPropertyName: string = toProperty instanceof Symbol ?
      typeof toProperty :
      toProperty;
```

This way every `Symbol` instance will have the `symbol` name, we can use `toProperty.toString()` instead but it's even easier just to remove Symbol altogether. Is there any particular reason you added it here?

Otherwise, it looks good for me for the first glance, so it will be good to go as we can fix these:
- I would like to avoid using `lodash` please try to remove it, 
- please follow the conventions in the repo, I don't like it either but for now please attach the decorator to the end of the `decorators.ts` file
- please implement this decorator the same way as the other decorators are written
- also as you mentioned, please rewrite the test to use mocha
- please rename your DTOs into more neutral names, like User, Admin, etc. Not everyone else is a Java dev, and they won't know what DTO refers to.

VinceOPS:
Hi @NoNameProvided 
Thanks for your feedback. Will work on this soon.
Regarding `string | symbol`, I wanted to add/handle `symbol` because this is how `PropertyDecorator` is defined. What do you think? If I'd like to keep it as a PropertyDecorator, what would be an acceptable solution for you?

vlapo:
@VinceOPS What is status of this PR? Are you still interested?

kenberkeley:
This feature is awesome! Nice to have!

kenberkeley:
Related: https://github.com/tannerntannern/ts-mixer/issues/15

kenberkeley:
Tips: `@InheritValidation` does not inherit `@Type`, have to add this manually...

tmtron:
could we make the decorator type-safe so that we don't accidentially have typos in the property name? e.g. maybe something like this:

```typescript
function InheritValidation<T>(fromClass: ClassType<T>, fromProperty?: keyof T)
```

kenberkeley:
This PR no longer works in v0.12+

NoNameProvided:
Hi @VinceOPS

Thanks for this contribution and sorry for the delayed response. Are still interested in this PR? If so can you rebase these changes to the current develop branch?

We have 
  - moved to Jest so you need to adapt your tests (it's straightforward)
  - added Prettier you need to format your code with `npm run prettier:fix`
  - move the files around a bit (split decorators into their own file for example)

After rebase  I will do a review in a timely manner.

Issue #64, opened 2017-02-09
ValidationSchema - constraints and options should not be required


wookieb:
Hi,
As in the question above.

"Constraints" shouldn't since many validators don't require it (like IS_EMPTY)
"Options" because you don't need to provide any options most of the time

I'm happy to fix it, just need your voice here.

Good job BTW.

NoNameProvided:
Fixed, sorry for the delay.

PR #163, opened 2018-01-30
fix: changed contraints and options to optional properties in Validat…

abinici:
Fix for https://github.com/typestack/class-validator/issues/64


NoNameProvided:
Thanks!

PR #170, opened 2018-02-16
Fixed case when validationArguments.value is not string

kemalmustafic:
When **validationArguments.value** is not string it tries to replace parts of message (labeled with $value) and produces error


harisb2012:
Had similar issue few days ago. Would be great to have this merged.

tms1337:
Same here.

PR #174, opened 2018-03-04
Add Table of Contents to README

cxcorp:
Navigating the documentation in the README was a bit laborious because there was no table of contents from which you could see what's included.

This commit adds a table of contents similar to that of [routing-controllers](https://github.com/typestack/routing-controllers/blob/37feaffa7e2cc45e1fffb1a7d1f472e258868e4f/README.md#table-of-contents).


NoNameProvided:
Thanks for the contribution! 🎉

cxcorp:
My pleasure! Thanks for the great work on the library!

PR #196, opened 2018-04-24
make IsOptional conditional

jbdemonte:


scyrencop (approved):

fr1sk:
Why is this one still not merged? I want to have optional property which will be optional only if some condition is true. eg. If I use `validateIf()` validation will be ignored if the field is passed, but I just want to control isOptional with the condition, not the whole validation.

NoNameProvided:
Hi @jbdemonte!

Thanks for this contribution and sorry for the delayed response. Can you rebase these changes to the current develop branch?

We have 
  - moved to Jest so you need to adapt your tests (it's straightforward)
  - added Prettier you need to format your code with `npm run prettier:fix`

After rebase I will review the pushed code in a timely manner.

NoNameProvided:
Superset by https://github.com/typestack/class-validator/pull/689.

Issue #203, opened 2018-05-23
ValidateNested without decorator is not working


zender:
Let's say we have the following validation schema:
`
// this will not work
export const taskValidationSchema: ValidationSchema = {
  name: 'TaskValidationSchema',
  properties: {
    'title': [
      {
        type: 'minLength',
        constraints: [3],
        groups: ['admin'],
      }
    ],
    'files': [
      {
        type: 'nestedValidation',
        each: true,
        groups: ['admin'],
      }
    ]
  }
};`

Validator is not aware of which validation schema to use for the nested property.

Here is my solution:

export const taskValidationSchema: ValidationSchema = {
  name: 'TaskValidationSchema',
  properties: {
    'title': [
      {
        type: 'minLength',
        constraints: [3],
        groups: ['admin'],
      }
    ],
    'files': [
      {
        type: 'nestedValidation',
        options: {'validationSchema': 'TaskFileValidationSchema'},
        each: true,
        groups: ['admin'],
      }
    ]
  }
};`
`

Would you accept PR?

zender:
#205

enrico-calzavara:
@NoNameProvided any news on this?

henrikra:
@rlataguerra Are you experiencing the same as I with `ValidateNested` decorator? Can you also check @NoNameProvided and @19majkel94

aviranhanan:
@NoNameProvided there is some estimation to fix this issue?

joseym:
😞  so I guess this is still a problem with no fix on the roadmap...

I'm trying to dynamically generate a validation schema based on the results of an xhr request, and building a class isn't possible...

NoNameProvided:
Marking this as blocked, as I want to start a discussion around the deprecation of the schema-based validation. I will CC the participants once I have created the issue.

PR #205, opened 2018-05-23
feat: add option to set validation schema in nested validator

zender:
references #203

### DIFF
--- a/src/validation/ValidationExecutor.ts
+++ b/src/validation/ValidationExecutor.ts
@@ -257,7 +257,12 @@ export class ValidationExecutor {
 
         metadatas.forEach(metadata => {
             if (metadata.type !== ValidationTypes.NESTED_VALIDATION) return;
-            const targetSchema = typeof metadata.target === "string" ? metadata.target as string : undefined;
+            let targetSchema: string;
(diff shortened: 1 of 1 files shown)


NoNameProvided:
I am not sure on this API. We should pass it as a constraint and modify the ValidateNested decorator to expect a class definition as the first parameter.

enrico-calzavara:
@NoNameProvided do you mean pass the nested schema name in the constraints array of the validation schema? And what about the class definition?

diosney:
@NoNameProvided Any updates on this? Please, this is a showstopper for me to using the library, since I do use Interfaces instead of classes, so the use of the decorators is not a fix for me.

19majkel94:
@diosney 
pure `abstract class` is effectively an `interface`, so you can use it with decorators and other interfaces might implement an abstract class

diosney:
@19majkel94 Thanks for the tip! I will check if it is possible for me to change from interface based type checking to abstract class based validation.

Thanks!

NoNameProvided:
Closing this as this won't get merged in its current state. The issue stays open for discussion.

Issue #202, opened 2018-05-23
upgrade validator.js


xujif:
to fix some bugs.

NoNameProvided:
Upgraded in 9122b4e and will be included in the next release.

PR #206, opened 2018-05-24
update validator.js to latest

elthrasher:
Validator 9.2 has a security vulnerability so would be good to update.
https://snyk.io/vuln/npm:validator:20180218
There are a couple of changes which caused tests to fail. validator.js no longer allows multiple dots in gmail addresses
https://github.com/chriso/validator.js/blob/master/validator.js#L215
They also changed the regexp for isNumber to support floating points
https://github.com/chriso/validator.js/commit/8ba0c299a75f2cf8862013b9a4bcaef46c0bd80a
I fixed the tests but this obviously changes how the library works a bit.
This solves https://github.com/typestack/class-validator/issues/202

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,11 +10,11 @@
       "integrity": "sha1-z6I7xYQPkQTOMqZedNt+epdLvuE=",
       "dev": true,
       "requires": {
-        "acorn": "^5.0.3",
-        "css": "^2.2.1",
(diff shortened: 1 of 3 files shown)


NoNameProvided:
Hi! 

Thanks for the contribution! It's all good with this PR but it has been superset by 9122b4e so I will close this without merge. Thanks again for your contribution!

PR #251, opened 2018-08-14
Fix custom message function signature typings

### LINKED ISSUE #250: Types: Wrong signature for custom error message function
I want to supply a custom message function to a decorator like this:

```ts
@IsString({ message: myCustomMessageFn })
myString: string;
```

Typescript says the [function signature](https://github.com/typestack/class-validator/blob/master/src/decorator/ValidationOptions.ts#L19) is:

```ts
((value?: any, constraint1?: any, constraint2?: any) => string)
```

but class-validator will actually pass an object that the fits [`ValidationArguments`](https://github.com/typestack/class-validator/blob/master/src/validation/ValidationArguments.ts#L5) as the first paramter.

I guess the this shoud actually be:
```ts
message?: string | ((validationArguments: ValidationArguments) => string);
```

hpohlmeyer:
Fixes #250

### DIFF
--- a/src/decorator/ValidationOptions.ts
+++ b/src/decorator/ValidationOptions.ts
@@ -1,3 +1,5 @@
+import {ValidationArguments} from "../validation/ValidationArguments";
+
 /**
  * Options used to pass to validation decorators.
  */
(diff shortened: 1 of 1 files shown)


NoNameProvided:
Thanks for your contribution!

PR #255, opened 2018-09-07
Create .npmignore

styfle:
I don't believe you need to publish the `js.map` files.
This will avoid publishing those files to npm along with everything else in your gitignore.

See [class-validator](https://packagephobia.now.sh/result?p=class-validator) size over time.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,8 @@
+build/
+node_modules/
+coverage/
+npm-debug.log
+*.iml
+.idea/
+
+*.js.map


19majkel94:
I don't ship `map` files, at least in latest releases.

And I'm using custom build script so npm ignore is not needed as package bundle have it's own directory.

Anyway, thanks for good intentions 😉

styfle:
I see several `.js.map` files in the latest version on npm

https://unpkg.com/class-validator@0.9.1/validation/

double checking

https://cdn.jsdelivr.net/npm/class-validator@0.9.1/validation/

19majkel94:
@styfle Sorry, I was thinking this PR is for `type-graphql` 😆

styfle:
But you might want to use `.js.map` files for development.

The point of this PR is that they are not necessary for other users who install `class-validator` which is why npm can ignore them during publishing.

styfle:
@19majkel94 Any update here?

PR #264, opened 2018-10-05
Expose default container instances to external containers

fernandolguevara:
Expose default container instances to external containers when useContainer is invoke

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -204,7 +204,7 @@ export class Gulpfile {
      */
     @SequenceTask()
     tests() {
-        return ["compile", "tslint", "coveragePost", "coverageRemap"];
+        return ["tslint", "compile", "coveragePost", "coverageRemap"];
     }
 
 }
--- a/src/container.ts
+++ b/src/container.ts
@@ -5,7 +5,7 @@ type DefaultContainer = {
     isEmpty(): boolean;
 };
 
-type Entry = { type: Function, object: any };
+export type Entry = { type: Function, object: any };
 
 /**
  * Container options.
@@ -26,7 +26,7 @@ export interface UseContainerOptions {
      * If the default container has registered instances this function allow to class-validator to expose the current state.
      * @param instances 
      */
-    registerExistingInstances(instances: Entry[]): void;
+    registerExistingInstances?: (instances: Entry[]) => void;
 }
 
 /**
--- a/src/container.ts
+++ b/src/container.ts
@@ -1,4 +1,12 @@
 
+type DefaultContainer = {
+    get<T>(someClass: { new(...args: any[]): T } | Function): T;
+    getInstances(): Entry[];
+    isEmpty(): boolean;
+};
+
+type Entry = { type: Function, object: any };
+
 /**
  * Container options.
  */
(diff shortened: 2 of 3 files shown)


fernandolguevara:
@19majkel94 review please =)

19majkel94:
What is the purpose of that feature? Why you can't just use TypeDI along with class-validator?

fernandolguevara:
@19majkel94 of course i can use type di or another IoC container, but this feature is to expose the internal configuration of class-validator.

run the next example  

```javascript
import { validate, useContainer, IsNotEmpty } from "../../src/index";

export class Customer {
    @IsNotEmpty()
    name: string;
}

const customer = new Customer();

validate(customer).then(result => {
    console.log('1 val', result);
});

const defaultContainer = new (class {
    private instances: any[] = [];
    get<T>(someClass: { new(...args: any[]): T }): T {
        let instance = this.instances.find(instance => instance.type === someClass);
        if (!instance) {
            instance = { type: someClass, object: new someClass() };
            this.instances.push(instance);
        }

        return instance.object;
    }

    set(instances: any[]): void {
        this.instances = [...instances, this.instances];
    }
})();

useContainer(defaultContainer, {
    registerExistingInstances: (instances) => {
        // defaultContainer.set(instances); // un-comment  
    }
});

validate(customer).then(result => {
    console.log('2 val', result);
});
```

if you import a class decorated before you call useContainer function, an instance of MetadataStorage is registered in the default container with the configuration of imported class, so its posible that in our systems exists two instances of MetadataStorage with partial configurations one created by the internal container of the library and one created by type di.

DavidBabel:
I like it :)

vlapo:
@satanTime could you please check this PR and https://github.com/typestack/class-validator/pull/265. I would like to hear your opinion on these changes.

satanTime:
Sure, I'll do that tonight.

satanTime:
Hi @vlapo,

Long story short: it has been solved by b57fef48f827432bd6f7df2027ce5111ccfc599e.

This issue is about the similar problem when we want to get access to the original metadata definition when we have different DI providers / contexts.
The author says:
> if you import a class decorated before you call useContainer function, an instance of MetadataStorage is registered in the default container...

Because now `MetadataStorage` is shared among whole applications through `getMetadataStorage()` and the `global.classValidatorMetadataStorage` variable, there is no need in the PR and it can be simply closed.

Let me know if you need more input from me.

vlapo:
Thank you @satanTime. I agree with you. Just want to know your opinion as you modify DI mechanism lately.

Feel free to reopen in case someone has different opinion.

fernandolguevara:
@satanTime @vlapo  
Hi guys, next time I hope you accept the contributions of your users, you could have accepted my pull request instead of copying the code that was in the PR

satanTime:
Hi @fernandolguevara, 

I think there's a misunderstanding, your PR was closed without copying. The issue was solved in a bit different way, like typeorm and routing-controllers do it.

Feel free to ping me if the fix doesn't work for you.
Then we could contribute to typeorm, routing-controllers and here with a better solution.

fernandolguevara:
@satanTime do you see the code on the PR #265 ? i think is the same code that yout commit here b57fef4 👎🏽

satanTime:
Ah, now I see what you mean. I can assure you my PR was copy of routing controllers implementation from here https://github.com/typestack/routing-controllers/commit/c263928d08ce3b41f843256f4a3a444d176f86d9.

vlapo:
@fernandolguevara Nobody copied your solution. I think @satanTime only had same idea (inspired by typeorm).

It was not my intention to accept his PR before yours. It was just my fault that I did not notice the similarity of the PRs. I'll be more careful next time.

PR #272, opened 2018-10-26
Automatically validating properties on assignment

NinnOgTonic:
Hi,

While I really like the annotation approach to further specify the possible values of variables in classes, much like a range of other classes.

We were considering including the class validator library in some projects, where there are a wide range of different DTOs we would like to verify.

In that connection, it felt rather bloated to have to manually invoke the validation every time a new object is created, as such it occurred to me that it would, at least for our use case, be possible to do all the validation entirely behind the scenes.

I know that there are edge cases, such as if a property is never assigned, and left as undefined, this wont be caught automatically -- but I recon that many like my use case is able to guarantee assignment to all relevant fields.

While I am not that deep into the codebase of the project, I have attempted to implement the automatic validation via property setters, along with adding an extra annotation which should target the class containing the properties with validator annotations attached!

Please let me know if you have any input.

### DIFF
--- a/sample/sample8-automatic-validation/Post.ts
+++ b/sample/sample8-automatic-validation/Post.ts
@@ -1,0 +1,39 @@
+import {Contains, IsInt, MinLength, MaxLength, IsEmail, IsFQDN, IsDate, ArrayNotEmpty, ArrayMinSize, ArrayMaxSize, IsEnum, AutoValidate} from "../../src/decorator/decorators";
+
+export enum PostType {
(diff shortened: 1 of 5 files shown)


vlapo:
Hi @NinnOgTonic. 
I want thank you very much for this PR and proposal of functionality. Since this PR is pretty old and no one else found this functionality as useful (no comments or likes) I will close this as we won't introduce this functionality in class-validator.

PR #273, opened 2018-11-01
feat: add @IsISO31661Alpha3 and @IsISO31661Alpha2 validators

neilime:
Validator exists for [validator.js](https://github.com/chriso/validator.js)

According to wikipedia : ISO_3166-1_alpha-3 allows a better visual association between the code and the country names than the two-letter alpha-2 codes
(https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3)


d-yokoi:
I appreciate if you could add alpha-2 as well :)

neilime:
@d-yokoi : Done

PR #275, opened 2018-11-01
Update IsDateString to match only entire ISO Date

vincent-chapron:
Regex checked only if ISO date was present in the string, it should check if the entire string is ISO Date


NoNameProvided:
Hi @vincent-chapron! 

Thanks for your contribution!

PR #282, opened 2018-11-09
feat: add isPort decorator

henrikra:
## Background
I have many validation classes where I want to validate ports and ended up doing this many times

```ts
@IsInt()
@Min(0)
@Max(65535)
port: number;
```

But as you notice it will get very repetitive  when I have to do this validation many for many properties.

Then I noticed validator.js already has validator for this: https://github.com/chriso/validator.js/#validators (search for isPort). So I ended up adding `IsPort` decorator.

## Result

Now I can just validate port like this
```ts
@IsPort()
port: number;
```

Also when I imported validator js correctly the code can use it types correctly. There were two mismatches with the types so I fixed them too. If you for example think that types in `isISBN` are wrong they should be fixed in validatorjs types repo :)

Tell me what do you guys think <3


henrikra:
@19majkel94 Care to check? :)

src/validation/Validator.ts
```diff
@@ -15,7 +16,7 @@ export class Validator {
     // Private Properties
     // -------------------------------------------------------------------------
 
-    private validatorJs = require("validator");
+    private validatorJs = validator;
```
vlapo: I am not sure about this line. We can update this later or maybe use validator.js as peer dependency but I have to check more info about it. Maybe we should use for now:
```
private validatorJs: ValidatorJS.ValidatorStatic = require("validator");
```

henrikra:
@vlapo Before merging this PR I created another to DefinitelyTyped https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36978/ because we are using their function which is not typed yet :P

henrikra:
@vlapo Now they merged my PR so can you merge this pr which includes the updates? https://github.com/typestack/class-validator/pull/383

After that PR is merged my PR will work :)

henrikra:
@vlapo Great! When you are going to make new release to npm?

vlapo:
Do not have access to npm yet and I want to resolve one big issue before release (tree shaking, packaging and package size). So I do not have an ETA. I hope !maybe! end of month. :)

henrikra:
Hmm okay. But wouldn't it make sense to do small releases as possible? Because if something breaks it easier to see what caused it :) VS if you released bunch of new fixes and features at the same time :P

vlapo:
Sure lets do this in small steps. https://www.npmjs.com/package/class-validator/v/0.10.0-rc.1 happy testing :) We jump from typescript@2.9.2 to 3.5.3. So we have to test this version on project with typescript < 3.0. I think there are some breaking changes with typings generated by typescript@3.

henrikra:
Excellent!

henrikra:
At least looks like `IsPort` is working correctly. Time to bumb to 0.10.? :D

PR #291, opened 2018-11-27
Added constraintCls field to ValidationSchema

rostyslav-diakiv-binary:
This change adds a possibility to use custom validation with ValidationSchema


icewind:
Can we get this merged? @pleerock @NoNameProvided 
Need custom validators using schemas.

vlapo:
Could you please add tests and docs?

NoNameProvided:
Blocking this PR before we discuss schema validation in the library. I want to deprecate that functionality as there are a lot of libraries which does JSON schema-based validation, we should focus on decorator based validation.

PR #310, opened 2019-01-25
feat: replace instanceof Promise and support Promise/A+ 

andrewwa-kainos:
This fixes a problem found wherein "Promise" instances from ES5 were not recognised as "instanceof Promise" from ES2015.


package-lock.json
```diff
@@ -10,11 +10,11 @@
...
-        "acorn": "5.7.1",
-        "css": "2.2.3",
-        "normalize-path": "2.1.1",
-        "source-map": "0.6.1",
-        "through2": "2.0.3"
+        "acorn": "^5.0.3",
+        "css": "^2.2.1",
+        "normalize-path": "^2.1.1",
+        "source-map": "^0.6.0",
+        "through2": "^2.0.3"
```
sabahirfan: do we need to bump these versions down?

timja:
@NoNameProvided Would you be able to take a look at this please when you have time? Thanks

vlapo:
I am ok with support Promise/A+ compliant thenable but why we need another dependency because of one function? Could you please create e.g. new `utils.ts` file with `isPromise` function?

andrewwa-kainos:
@vlapo Done

src/utils.ts
```diff
@@ -0,0 +1,8 @@
+// https://github.com/TylorS/typed-is-promise/blob/abf1514e1b6961adfc75765476b0debb96b2c3ae/src/index.ts
+
+export function isPromise<T>(p: any): p is Promise<T>;
+export function isPromise(p: any): p is Promise<any>;
+
+export function isPromise<T>(p: any): p is Promise<T> {
```
vlapo: Are these function overloads necessary. 
Define function as `export function isPromise<T = any>(p: any): p is Promise<T = any> {` should be enough.

andrewwa-kainos: I'm far from familiar with TypeScript generics, but this looks reasonable. I had just copied it from the original library. Shall action.

vlapo:
Thank you for your contribution :)

PR #311, opened 2019-01-28
ROC-4880 instanceof promise failure

andrewwa-kainos:


andrewwa-kainos:
Apologies. Meant to be local.

Issue #312, opened 2019-02-02
ValidateNested, unhandled exception when provided array contains null


golozubov:
Hi!

I'm getting TypeError when validating array of objects that contains nulls among another objects.
Let's consider the next example:

```ts
class MySubClass {
    @MinLength(5)
    name: string;
}

class MyClass {
    @ValidateNested()
    mySubClasses: MySubClass[];
}

const model = new MyClass();
model.mySubClasses = [null, new MySubClass()];
```

Here you can see `null` in `mySubClasses` array. Looks legit, right?

But invocation of `validator.validate(model)` causes the next error (I've succeeded reproducing this in tests):

```
TypeError: Cannot read property 'constructor' of null
    at ValidationExecutor.execute (build/compiled/src/validation/ValidationExecutor.js:43:88)
    at /Users/user/Develop/class-validator/build/compiled/src/validation/ValidationExecutor.js:218:27
    at Array.forEach (<anonymous>)
    at /Users/user/Develop/class-validator/build/compiled/src/validation/ValidationExecutor.js:215:23
    at Array.forEach (<anonymous>)
    at ValidationExecutor.nestedValidations (build/compiled/src/validation/ValidationExecutor.js:209:19)
    at /Users/user/Develop/class-validator/build/compiled/src/validation/ValidationExecutor.js:82:19
    at Array.forEach (<anonymous>)
    at ValidationExecutor.execute (build/compiled/src/validation/ValidationExecutor.js:62:39)
    at Validator.coreValidate (build/compiled/src/validation/Validator.js:63:18)
    at Validator.validate (build/compiled/src/validation/Validator.js:72:21)
    at Context.<anonymous> (build/compiled/test/functional/nested-validation.spec.js:177:26)

```

Desired behavior: don't fail on `null`, just skip it and continue validating the rest elements of array.

golozubov:
Also I'm getting the next error with `whitelist: true`
```
TypeError: Cannot delete property '0' of [object String]
    at /Users/user/Develop/class-validator/build/compiled/src/validation/ValidationExecutor.js:106:89
    at Array.forEach (<anonymous>)
    at ValidationExecutor.whitelist (build/compiled/src/validation/ValidationExecutor.js:106:38)
    at ValidationExecutor.execute (build/compiled/src/validation/ValidationExecutor.js:60:18)
    at /Users/user/Develop/class-validator/build/compiled/src/validation/ValidationExecutor.js:218:27
    at Array.forEach (<anonymous>)
[...]

PR #313, opened 2019-02-02
fix: validation of nulls in nested array with `ValidateNested` throws runtime error 

golozubov:
Possible fix of #312 
I would appreciate any ideas or feedback.

### DIFF
--- a/src/validation/ValidationExecutor.ts
+++ b/src/validation/ValidationExecutor.ts
@@ -40,7 +40,7 @@ export class ValidationExecutor {
     // -------------------------------------------------------------------------
 
     execute(object: Object, targetSchema: string, validationErrors: ValidationError[]) {
-        if (object === undefined || object === null) {
+        if (object !== Object(object)) {
             return;
         }
         /**
--- a/src/validation/ValidationExecutor.ts
+++ b/src/validation/ValidationExecutor.ts
@@ -40,6 +40,9 @@ export class ValidationExecutor {
     // -------------------------------------------------------------------------
 
     execute(object: Object, targetSchema: string, validationErrors: ValidationError[]) {
+        if (object === undefined || object === null) {
+            return;
+        }
         /**
          * If there is no metadata registered it means possibly the dependencies are not flatterned and
          * more than one instance is used.
--- a/test/functional/nested-validation.spec.ts
+++ b/test/functional/nested-validation.spec.ts
@@ -138,4 +138,31 @@ describe("nested validation", function () {
 
     });
 
+    it("should not validate nulls in nested array", () => {
+
+        class MySubClass {
+            @MinLength(5)
+            name: string;
+        }
+
+        class MyClass {
+            @ValidateNested()
+            mySubClasses: MySubClass[];
+        }
+
+        const model = new MyClass();
+        model.mySubClasses = [null, new MySubClass()];
+        model.mySubClasses[1].name = "my";
+
+        return validator.validate(model).then(errors => {
+            expect(errors[0].target).to.equal(model);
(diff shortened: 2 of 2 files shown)


diosney:
Any updates on this?

golozubov:
@pleerock @NoNameProvided Could you please take a look at this fix when you'll have a chance?

src/validation/ValidationExecutor.ts
```diff
@@ -40,6 +40,9 @@ export class ValidationExecutor {
     // -------------------------------------------------------------------------
 
     execute(object: Object, targetSchema: string, validationErrors: ValidationError[]) {
+        if (object !== Object(object)) {
```
amirsaber: I don't think this is the right place to add this. I think this should add the fix to `nestedValidations()`

test/functional/nested-validation.spec.ts
```diff
@@ -138,4 +138,57 @@ describe("nested validation", function () {
...
+        model.mySubClasses[1].name = "my";
+
+        return validator.validate(model).then(errors => {
+            expect(errors[0].target).to.equal(model);
+            expect(errors[0].property).to.equal("mySubClasses");
+            expect(errors[0].children.length).to.equal(1);
+            expect(errors[0].children[0].children.length).to.equal(1);
+
+            const subError = errors[0].children[0].children[0];
+            subError.constraints.should.be.eql({minLength: "name must be longer than or equal to 5 characters"});
```
amirsaber: Shouldn't we show a message regarding the null item instead of this?

saberistic (changes requested):

test/functional/nested-validation.spec.ts
golozubov: @amirsaber I don't think so. The original problem was related to unhandled exception in class-validator module. And this PR just fixes this issue (makes class-validator don't regard nulls as objects in nested array). It's just input data filtering.
According to https://github.com/typestack/class-validator#validating-nested-objects ValidateNested() will only validate nested object if it's present. But as for me, checking for array content should be performed in another validator, such as `@IsDefined({ each: true })`.
Imho the situation when you'll have an array that can contain either nulls or objects with specified format isn't impossible.

src/validation/ValidationExecutor.ts
golozubov: @amirsaber Let's consider the next example: `validator.validate(null)`.
It will throw `TypeError: Cannot read property 'constructor' of null` without this fix.
It seems to be incorrect behavior. So we shouldn't place this check only to `nestedValidations()`.

test/functional/nested-validation.spec.ts
vlapo: We should add note about skipping non-object values in `@ValidateNested()` to the docs. Also small recommendation about using `@IsObject()` or `@IsDefined()` to handle this case.

src/validation/ValidationExecutor.ts
vlapo: I agree we should handle also cases like `validator.validate(null)`. But I am a little bit worried about ending this case without any error. I think in this case we should throw an error with some message about invalid input value. What do you think @golozubov, @amirsaber

NoNameProvided:
I Will take a look at this, but I have to think this over, where to put this functionality.

molinajulian:
Any updates about this?

amirsaber:
I believe we need a rebase before re-evaluating. Conflicts should be resolved.

NoNameProvided:
Closing this as out of date. The reproduction from the issue returns the expected validation errors.

PR #329, opened 2019-03-12
build: update validator

d-yokoi:
I would like to use the `disallow_auth` option of `isURL` validation, which was added recently.


vlapo:
Close as #378 is merged

Issue #308, opened 2019-01-24
Null value passes when skipMissingProperties set to true


iveselin:
When skipMissingProperties is set to true, if property has null value, it will not validate with IsNotEmpty decorator and null value will be forwarded to controller/service/db repo etc. and cause a crash
[Source of error](https://github.com/typestack/class-validator/blob/master/src/validation/ValidationExecutor.ts#L100)

### Example
Json input
```javascript
{
    "startDate":null 
}
```

Part of UpdateProjectDto
```javascript
 @IsNotEmpty()
 @IsDateString()
 readonly startDate: Date;
```

Nest.js framework controller example of code which will receive poData with startDate set to null and crash at DB query...
```javascript
@Patch('/:id')
@UsePipes(new ValidationPipe({ skipMissingProperties: true }))
    updateById(@Param('id', new ParseIntPipe()) id, @Body() poData: UpdateProjectDto) {
        //#update route
        return this.projectsService.updateById(id, poData);
   }
```

nigelthinksprint:
any idea why this might be anyone?

zhenwenc:
Just search the source code :-) 

```
// ValidationExecutor#execute
...
            if ((value === null || value === undefined) && this.validatorOptions && this.validatorOptions.skipMissingProperties === true) {
                return;
            }
...
```

https://github.com/typestack/class-validator/blob/master/src/validation/ValidationExecutor.ts#L100

jumojer:
I'm having a similar issue.
Why would we consider `null` as equivalent to `undefined` in this case?
I believe this behavior breaks partial object validation for non-nullable fields.

For example:
```javascript
    @IsString()
    @IsNotEmpty()
    @IsDefined()
    name: string;
```
Those two objects would both pass validation when the `skipMissingProperties` is applied
```javascript
validate({}, { skipMissingProperties: true})
validate({ "name": null }, { skipMissingProperties: true})
```
Which, in my case, triggers a database `Column 'name' cannot be null` error.

Wouldn't it make more sense if  `skipMissingProperties` and the `IsDefined` decorators relied solely on whether the property is, well, defined instead of also checking if it is `null`?

thomas-ama:
Some properties of my model can be `null`, some can't. None can be `undefined`.
In case of `PUT` / `PATCH` HTTP requests, I'm using `{ skipMissingProperties: true}` to validate the body of my request because it contains only updated properties.

However:
- `{ skipMissingProperties: true}` doesn't validate `null` values
[...]

PR #353, opened 2019-05-10
Skip undefined properties

thomas-ama:
PR for issue #308

### DIFF
--- a/src/validation/ValidatorOptions.ts
+++ b/src/validation/ValidatorOptions.ts
@@ -9,7 +9,7 @@ export interface ValidatorOptions {
     skipUndefinedProperties?: boolean;
 
     /**
-     * If set to true than validator will skip validation of all properties that are missing in the validating object.
+     * If set to true than validator will skip validation of all properties that are null or undefined in the validating object.
      */
     skipMissingProperties?: boolean;
 
--- a/src/validation/ValidatorOptions.ts
+++ b/src/validation/ValidatorOptions.ts
@@ -4,12 +4,12 @@
 export interface ValidatorOptions {
 
     /**
-     * If set to true than validator will skip validation of all properties that are undefined in the validating object.
+     * If set to true then validator will skip validation of all properties that are undefined in the validating object.
      */
     skipUndefinedProperties?: boolean;
 
     /**
-     * If set to true than validator will skip validation of all properties that are null or undefined in the validating object.
+     * If set to true then validator will skip validation of all properties that are null or undefined in the validating object.
      */
     skipMissingProperties?: boolean;
 
--- a/src/validation/ValidatorOptions.ts
+++ b/src/validation/ValidatorOptions.ts
@@ -4,6 +4,11 @@
 export interface ValidatorOptions {
 
     /**
+     * If set to true than validator will skip validation of all properties that are undefined in the validating object.
+     */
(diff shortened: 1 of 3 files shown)


thomas-ama:
PR for issue #308

thomas-ama:
I don't understand why tests are failing on Node.js stable but not on LTS. Any ideas?

henrikra:
This is something we really need also! Have you pinged any maintainers so we get this in?

thomas-ama:
@NoNameProvided @19majkel94

src/validation/ValidatorOptions.ts
```diff
@@ -3,6 +3,11 @@
  */
 export interface ValidatorOptions {
 
+    /**
+     * If set to true than validator will skip validation of all properties that are undefined in the validating object.
+     */
+    skipUndefinedProperties?: boolean;
+
     /**
      * If set to true than validator will skip validation of all properties that are missing in the validating object.
```
henrikra: Should you also update this documentation about `skipMissingProperties`?

thomas-ama: What do you mean? `skipMissingProperties` still checks `null` and `undefined` values.
Do you want to replace

> that are missing

with

> that are null or undefined

?
Or do you want to have a new option `skipNullProperties`?

henrikra: Yes replace missing documentation with `that are null or undefined` or something similar :)

henrikra:
Also if you remove this line https://github.com/typestack/class-validator/blob/master/.travis.yml#L3 then the test will pass becase stable is getting node 12 which is not actually stable :P If you google the error on travis logs you will get lot of cases like this https://www.google.com/search?q=ReferenceError%3A+primordials+is+not+defined&oq=ReferenceError%3A+primordials+is+not+defined&aqs=chrome..69i57j69i58j69i60.220j0j7&sourceid=chrome&ie=UTF-8

src/validation/ValidatorOptions.ts
thomas-ama: Done

thomas-ama:
Thanks @henrikra, let's say tests are okay then.

src/validation/ValidatorOptions.ts
```diff
@@ -4,7 +4,12 @@
 export interface ValidatorOptions {
 
     /**
-     * If set to true than validator will skip validation of all properties that are missing in the validating object.
+     * If set to true than validator will skip validation of all properties that are undefined in the validating object.
```
henrikra: Both of these lines have typo :D `than` to `then`

henrikra:
This looks golden to me ⭐️

AlashaHonore:
Hello, do you have any idea when this flag will be available ? NEED 😄

henrikra:
Lets merge this!!

vlapo:
@thomas-ama could you please rebase and add also `skipNullProperties` option? We should deprecate `skipMissingProperties` in future and use `skipNullProperties` and `skipUndefinedProperties` instead.

ruscon:
@thomas-ama ping

vlapo:
Close in favore of https://github.com/typestack/class-validator/pull/414. @thomas-ama thank you for your effort.

PR #355, opened 2019-05-23
Fix IsEmail validation failure with underscores

garlou:


vlapo:
Close as #378 is merged

PR #365, opened 2019-06-09
feat: allow validate Map/Set

Tarik02:


src/validation/ValidationExecutor.ts
```diff
@@ -270,6 +270,25 @@ export class ValidationExecutor {
...
+                value.forEach((subValue: any) => {
+                    const validationError = this.generateValidationError(value, subValue, index.toString());
+                    errors.push(validationError);
+
+                    this.execute(subValue, targetSchema, validationError.children);
+
+                    ++index;
+                });
+
+            } else if (value instanceof Map) {
```
vlapo: Looks like duplicate code. Should be in first statement:
```ts
if (value instanceof Array || value instanceof Map) {
```

Tarik02: As [you can see](https://travis-ci.org/typestack/class-validator/jobs/559511528), it is not the best idea to merge these calls...

PR #44, opened 2016-10-16
Validate API - promise rejection

19majkel94:
Implements #40, fixes #43.

PR #370, opened 2019-06-18
Add example for .validateOrReject usage

vvscode:


vvscode:
And in general - I see no documentation for `Validator` class. My example is described in tests https://github.com/typestack/class-validator/pull/44/files#diff-27b51e6f3099f1a9738d41affcf20acdR36 but it will be nice to see it in docs

vvscode:
@amirsaber  any ideas about why the test failed?

vlapo:
Why not use only `validateOrReject()` function? We should prefer one way in documentation.

Tests failed because of gulp. Fixed in latest source. Please rebase.

vlapo:
Also we have existing PR about validateOrReject https://github.com/typestack/class-validator/pull/320/files

vvscode:
@vlapo  I just didn't find a `valideOrReject()` in docs, and this one is first I discovered according tests

vlapo:
@vvscode thnx for PR but I think we should accept https://github.com/typestack/class-validator/pull/320. It also update `docs/basics/validating-objects.md` and I think we should use `validateOrReject` root function as preferred in docs.

PR #359, opened 2019-05-31
feat(IsDecimal): implement 'IsDecimal' from validatorjs

nikvaessen:

PR #381, opened 2019-07-19
feat: add support for maxDecimalPlaces on IsNumber

jasperblues:


vlapo:
validator.js has `decimal_digits` option in `isDecimal` validator. We have a PR https://github.com/typestack/class-validator/pull/359 about adding `IsDecimal`.

Issue #266, opened 2018-10-12
fix: expose locale support for isAlpha and IsAlphanumeric decorator


TheNoim:
`isAlpha` and `isAlphanumeric ` are not useful in an non english country if they don't support the locals argument like validator.js does. 
This package is based on validator.js, so why are locals not supported?
```typescript

import { IsAlphanumeric } from 'class-validator';

export class Target {
    @IsAlphanumeric('de-DE')
    name: string;
}

```

This needs to be possible.

l2ubin:
Looking at the validator source code. It has a set of regex for all locales which it compares it with to determine a valid alpha and alphanumeric while in case of here, it is just  comparing using instanceof. I would be happy to code this feature

vlapo:
It is simple fix. Feel free create PR. Check e.g. `isMobilePhone` validator. Its a same principe.

l2ubin:
sure. I will work on it at my free time and create a pr

l2ubin:
added that on pr [#406](https://github.com/typestack/class-validator/pull/406)

rubiin:
@vlapo @NoNameProvided  I have added the fix

PR #384, opened 2019-07-24
refactor: use unknown type for validation functions inputs

henrikra:
I have refactored all `value` parameters correctly to `unknown` type. In short `unknown` type is like `any` but more strict because it forces you to make validation to it.

Before this PR lot of `value` looked like they have unneeded validation. For example:

```ts
isAscii(value: string): boolean {
  return typeof value === "string" && this.validatorJs.isAscii(value);
}
```

In this case it looks odd that why we have to validate that `value` is a string since we are telling TS that it is already a string. Correct way of doing this is like here: 

```ts
isAscii(value: unknown): boolean {
  return typeof value === "string" && this.validatorJs.isAscii(value);
}
```

This is correct since we cannot pass `value` to `this.validatorJs.isAscii` before we validate that the `value` is string 👍 

The side effect of this changes is that I might have caught some bugs. For example in `isNumber` we never checked that the `value` is number since you cannot pass anything you want to `Number.isNaN`. But we are checking that and are sure that everything works as expected :)

More info about `unknown` type here https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

### DIFF
--- a/src/validation/Validator.ts
+++ b/src/validation/Validator.ts
@@ -355,21 +355,21 @@ export class Validator {
     /**
     * Checks if a given value is a latitude.
     */
-    isLatLong(value: any): boolean {
-        return this.validatorJs.isLatLong(value);
+    isLatLong(value: unknown): boolean {
+        return typeof value === "string" && this.validatorJs.isLatLong(value);
     }
 
     /**
     * Checks if a given value is a latitude.
     */
-    isLatitude(value: any): boolean {
+    isLatitude(value: unknown): boolean {
         return (typeof value === "number" || this.isString(value)) && this.isLatLong(`0,${value}`);
     }
 
     /**
     * Checks if a given value is a longitude.
     */
-    isLongitude(value: any): boolean {
+    isLongitude(value: unknown): boolean {
         return (typeof value === "number" || this.isString(value)) && this.isLatLong(`${value},0`);
     }
 
@@ -708,7 +708,7 @@ export class Validator {
      * Checks if the value is valid Object.
      * Returns false if the value is not an object.
      */
-    isObject(value: any): boolean {
+    isObject(value: unknown): value is object {
         return value != null && (typeof value === "object" || typeof value === "function") && !Array.isArray(value);
     }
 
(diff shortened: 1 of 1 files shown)


henrikra:
@vlapo Do you have time to check this out? Tell me if you need some more info :)

vlapo:
Firstly I have to check this (for me) new `unknown` type. Dont worry I have this on my list. 
But we have more important tasks to done. Check new milestone https://github.com/typestack/class-validator/milestone/3 if you want more contribute :)

henrikra:
Hmm okay. Would it make sense to make new release always when new fix or feature is in master? Because if you want all of those in 0.10.1 it takes a lot of time and has many reasons for it to break instead of releasing in small increments :)

vlapo:
Version name `0.10.1` is only orientational. We can release 0.10.1 with fixes before all issues are resolved. I just want to pick some issues and make a list. But feel free to suggest issues to be included in the next release. I am open to all suggestions, contributions and helps.

henrikra:
Hmm well lets see: for example this issue https://github.com/typestack/class-validator/issues/266 is not a fix it is a new feature which didn't exist before so I would fix that and then make new release and the new version should be 0.11.0 (https://semver.org/) because we have added new functionality NOT fixed :)

After that I would keep going like this that make fix/feature -> release -> repeat. This way it is easy to see why something breaks IF something breaks :)

vlapo:
Sure I understand but we are still in `0.x` phase of versioning right now. I want to reserve increment of minor version number for cases when we have to break API (e.g. in our case updating validator.js library is almost every time breaking change). Hope we will reach `1.0` milestone soon and after that use semantic versioning.

To #266 - I think this is bug since `validator.js` implement this argument from beginning and `class-validator` not. But I agree it could be also adding new argument and support new feature/enhancement.

henrikra:
@vlapo Have you considered this pr? :D

henrikra:
@vlapo I really think this PR would help the code base a lot but what do you think should I close it someone else can do it later or should we review this? :D

vlapo:
Hey @henrikra sorry for such a delay. This change looks reasonable to me too. I hope we will also enable `strict` flag for TS soon.
Lets merge it and we will see. Could you please rebase and update also new validators?

henrikra:
@vlapo Now I have merged master and also added unknown to new functions!

And yeah you guys should definitely should concentrate on getting strict flag on 👍

vlapo:
Thank you for your effort!

PR #404, opened 2019-08-19
fix: add default message for isPort validator

l2ubin:
isPort which was used to check whether a given number was actaully a port was missing the default message thrown in case of falsy values.


vlapo:
Thank you :)

l2ubin:
I would like to help out with other pull requests as wel

vlapo:
Sure check issues or milestones and create PRs :)

l2ubin:
sure. On it

PR #406, opened 2019-08-20
fix: add locale parameter for isAlpha and isAlphanumeric validators

rubiin:
This commit adds the feature to add locale to both isAlpha and isAlphanumeric with default as en-US

### DIFF
--- a/src/validation/Validator.ts
+++ b/src/validation/Validator.ts
@@ -495,7 +495,7 @@ export class Validator {
      * Checks if the string contains only letters (a-zA-Z).
      * If given value is not a string, then it returns false.
      */
-    isAlpha(value: string, locale: ValidatorJS.AlphaLocale = "en-US"): boolean {
+    isAlpha(value: string, locale?: ValidatorJS.AlphaLocale): boolean {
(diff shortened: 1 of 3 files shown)


src/validation/Validator.ts
```diff
@@ -495,16 +495,16 @@ export class Validator {
      * Checks if the string contains only letters (a-zA-Z).
      * If given value is not a string, then it returns false.
      */
-    isAlpha(value: string): boolean {
-        return typeof value === "string" && this.validatorJs.isAlpha(value);
+    isAlpha(value: string, locale: ValidatorJS.AlphaLocale = "en-US"): boolean {
```
vlapo: Please leave locale parameter without default value. `validator.js` lib will use "en-US" by default. Same for isAlphanumeric.

l2ubin:
will do

l2ubin:
I have also fixed the number options on the isnumeric string on the same pr

l2ubin:
@vlapo I have addressed your changes.

vlapo:
Thank you for contribution.

rubiin:
Anytime

Issue #287, opened 2018-11-17
@IsMilitaryTime has no default message


bwegrzyn:
It seems like @IsMilitaryTime validator does not have a default message. I can fix it by providing a default message every time I use the decorator, but it would be nice to provide a default message.

PR #411, opened 2019-08-22
fix: add default message for isMilitaryTime validator

Kiliandeca:
Hello, this should fix issue #287 
It add a default message for the isMilitaryTime validator


vlapo:
Thank you :)

Issue #417, opened 2019-09-13
IsPhoneNumber does not match E.164


ihorTymofieiev:
# Description
Hi, I use the class-validator library to validate request data on a Node.js app. I have problems with validating phone numbers with intl. prefix (use 'ZZ' as a region)!

# Example
```
class PhoneNumberWithIssue {
  @IsNotEmpty({ message: 'ERROR_EMPTY_PHONE_NUMBER_FIELD' })
  @IsPhoneNumber('ZZ', { message: 'ERROR_INVALID_PHONE_NUMBER' })
  phoneNumber!: string;
}

const invalidPhoneNumber = new PhoneNumberWithIssue();
invalidPhoneNumber.phoneNumber = "+380981111111++++++++++=====++++=+";
validate(invalidPhoneNumber).then(errors => {
    if (errors.length > 0) {
        console.log("validation failed. errors: ", errors);
    } else {
        console.log("validation succeed"); // Validation is succeeded, but the phone number is not valid!
    }
});
```

# Dependencies

- `"class-validator": "^0.10.0"`

vlapo:
Looks like this is the way how to libphonenumber library parse strings. From javadocs:
> Parses a string and returns it as a phone number in proto buffer format. The method is quite lenient and looks for a number in the input text (raw input) and does not check whether the string is definitely only a phone number. To do this, it ignores punctuation and white-space, as well as any text before the number (e.g. a leading "Tel: ") and trims the non-number bits. It will accept a number in any format (E164, national, international etc), assuming it can be interpreted with the defaultRegion supplied. It also attempts to convert any alpha characters into digits if it thinks this is a vanity number of the type "1800 MICROSOFT".

So the question is. Is there any possible option how to do more strict parse? I dont think so https://github.com/typestack/class-validator/pull/421/ is good way how to handle this problem E164 is not only valid format.

ihorTymofieiev:
Hmm, then #421  is not a solution. I will try to make a more universal validation.

vlapo:
There are plenty of ports https://github.com/google/libphonenumber library to another programming languages. I hope someone have a similar issue. Maybe check official issue tracker for https://github.com/google/libphonenumber first. Some solution could be there. :) I will try check this too.

ihorTymofieiev:
https://github.com/google/libphonenumber ignores all characters after the phone number. For them, this is normal behavior. You can find notes about this in their documentation and as a comment in the source code.

rubiin:
[...]

PR #421, opened 2019-09-21
Add check for the IsPhoneNumber validator function to match E.164

ihorTymofieiev:
I found the problem with validating phone numbers. (issue #417 )

This issue occurs due to the peculiarity of the validation of the phone number in `google-libphonenumber`. The `parseAndKeepRawInput` method ignores the contents of the line after the phone number. At the same time, the `isValidNumber` method - checks the already processed string and not the one that was originally (`raw_input`).
In this solution, a regular expression was added to check for matching with the `E.164` standard.


NoNameProvided:
Closing as out-of-date as we have switched libraries.

Issue #387, opened 2019-07-26
feature: flag for `@IsDateString()` decorator to validate date without time


RomanSolianik:
It`s more feature request.
I think it would be nice to have ability validate date string without time using @IsDateString() decorator. For example, '2000-07-26', not '2000-07-26T00:00:01.967Z'. Cannot find how to do that not using @Matches() decorator.

vlapo:
Feel free to contribute :)

l2ubin:
I agree. There should be two different functions one for checking isostring and another for plain date string

l2ubin:
I can work on it if you like , also please let me know of steps for contributing

l2ubin:
@vlapo I think this needs further discussion what say. Please read the above comments too

rubiin:
@vlapo @RomanSolianik  I think it still needs some discussion

rubiin:
@vlapo i am taking this one

jerryfreshworks:
I create custom one.

```ts
import { registerDecorator, ValidationOptions } from 'class-validator';

export function IsOnlyDate(validationOptions?: ValidationOptions) {
  return function(object: Object, propertyName: string) {
    registerDecorator({
      name: 'IsOnlyDate',
      target: object.constructor,
      propertyName: propertyName,
      constraints: [],
      options: {
        message: 'Please provide only date like 2020-12-08',
        ...validationOptions,
      },
      validator: {
        validate(value: any) {
          const regex = /([12]\d{3}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01]))/;
          return typeof value === 'string' && regex.test(value);
        },
      },
    });
  };
}
```

dtmzr:
@jerryfreshworks as a little note - the provided validator passes `2020-02-31` as a correct date so it's rather a format validator than actually validating if the date exists.

kaisskander:
anything ?

talski:
for date format only you can use Matches validator
```js
@Matches(/^\d{4}(\/)(((0)[0-9])|((1)[0-2]))(\/)([0-2][0-9]|(3)[0-1])$/i, {
    message: "$property must be formatted as yyyy-mm-dd"
})
```

RomanSolianik:
@talski 
of course we can use regular expressions.. basically we can use them for all validation, but I think the point of the creation of this library was to simplify this process

rubiin:
> 2000-07-26

[...]

Issue #412, opened 2019-09-03
isDateString does not follow ISO-8601 ?


jacktuck:
Date strings with negative TZ offsets are not considered valid with current regex

Think it's missing negative offsets and handling single-digit hours
`2016-07-16T19:20:30+5:30` no match
`2016-07-16T19:20:30+05:30` match

`2019-09-03T20:16:24-01:00` no match
`2019-09-03T20:16:24+01:00` match

I *think* they should all match as valid iso strings, not too sure
Current regex testing those strings - https://regex101.com/r/RzcfrV/2/

Hapijs's joi (2.5M weekly downloads) is using the following regex 
```
^(?:[-+]\d{2})?(?:\d{4}(?!\d{2}\b))(?:(-?)(?:(?:0[1-9]|1[0-2])(?:\1(?:[12]\d|0[1-9]|3[01]))?|W(?:[0-4]\d|5[0-2])(?:-?[1-7])?|(?:00[1-9]|0[1-9]\d|[12]\d{2}|3(?:[0-5]\d|6[1-6])))(?![T]$|[T][\d]+Z$)(?:[T\s](?:(?:(?:[01]\d|2[0-3])(?:(:?)[0-5]\d)?|24\:?00)(?:[.,]\d+(?!:))?)(?:\2[0-5]\d(?:[.,]\d+)?)?(?:[Z]|(?:[+-])(?:[01]\d|2[0-3])(?::?[0-5]\d)?)?)?)?$
```

https://github.com/hapijs/joi/blob/e3e7626b39625c38a71f3b06665142ec0a29bac6/lib/common.js#L13
https://regex101.com/r/99Uj49/1/

Happy to make a PR to switch to Hapi's regex if it's correct.

vlapo:
Hi @jacktuck,

We are open to PR with fix and of course do not forget tests.

rubiin:
@vlapo  however validatorjs seems to have another regex https://github.com/validatorjs/validator.js/blob/master/src/lib/isISO8601.js

vlapo:
So why we have `isDateString` and also `isISO8601` validators? :) Looks like `isDateString` should be just alias for `isISO8601`. @rubiin thnx for the point us to this.

@jacktuck what do you think?

jacktuck:
@vlapo I must have missed the isISO8601 validator. Agree that isDateString should be an alias for isISO8601. Unless there is some important distinction that mean we'd want both.

rubiin:
[...]

PR #429, opened 2019-10-02
Feat: Added support for validation of date only in datesting

rubiin:
this pr is a reference to https://github.com/typestack/class-validator/issues/387 . Now we can use isDateString to validate datestring with dates only like "2011-10-19"


rubiin:
@vlapo

vlapo:
This PR is not necessary. Please check https://github.com/typestack/class-validator/issues/412

Issue #428, opened 2019-10-02
Trying to validate a Map but it doesn't seem to work


larsbs:
Doing exactly like in the provided example for Map validation doesn't work.

My code:
```javascript
class EventDto {
  @IsInt({ each: true })
  readonly clicks: Map<string, number>;
}
```

The above code won't trigger any errors when receiving: `{ myKey: 'a string' }`

PR #430, opened 2019-10-02
feat: support ES6 Map and Set for regular validators with each option

vlapo:
Looks like we do not support `each` option for Map and Set (https://github.com/typestack/class-validator/issues/428)

This PR add support Map/Set for regular validators and custom validators `@Validate`.

Also small unit tests added.

Close https://github.com/typestack/class-validator/issues/428


vlapo:
@henrikra and @rubiin as active contributors please review. All reviews are welcome.

rubiin:
Had a look at the code, doesn't seem to have any bugs so far @vlapo

Issue #440, opened 2019-10-25
Please use `zh-HK` to replace `en-HK`.


liuzhaowei55:
https://github.com/typestack/class-validator/blob/a98f5dd3e5693f112837f81befc600c66ba49885/src/decorator/decorators.ts#L972

https://github.com/validatorjs/validator.js/pull/622

vlapo:
What exactly is problem with `en-HK`?

liuzhaowei55:
> What is a problem with `en-HK`?

`zh-HK` can alo work very well
https://github.com/validatorjs/validator.js/blob/b5a1d1f23fcdd24e3ca9d6090fdc0a06fcc23d55/src/lib/isMobilePhone.js#L94

`en-HK`  is not supported in the locale
https://en.wikipedia.org/wiki/Locale_(computer_software)

政治正确 (Politically correct)

vlapo:
> > What is a problem with `en-HK`?
> 
> `zh-HK` can alo work very well
> [validatorjs/validator.js:src/lib/isMobilePhone.js@`b5a1d1f`#L94](https://github.com/validatorjs/validator.js/blob/b5a1d1f23fcdd24e3ca9d6090fdc0a06fcc23d55/src/lib/isMobilePhone.js#L94)

Ok but it is not a problem.

> `en-HK` is not supported in the locale
> [en.wikipedia.org/wiki/Locale_(computer_software)](https://en.wikipedia.org/wiki/Locale_(computer_software))

Do not see any info about not supporting `en-HK` in locale strings

> 政治正确 (Politically correct)

Please do not bring politics into open-source.

I do not see any problem neither with `en-HK` or `zh-HK` as `isMobilePhone` accepts both of them.

liuzhaowei55:
@vlapo 
I know `en-HK` still run very well, bug it no longer the standard.
https://en.wikipedia.org/wiki/Locale_(computer_software) 
And it is easy to cause misunderstanding, this is just a comment, why not change it to a standard?

vlapo:
I have few reservations:
 - First at all - you didnt show any proof about non-standard locale `en-HK`. I searched some registers and found `en-HK` as pretty used:
   - http://demo.icu-project.org/icu-bin/locexp?d_=en&_=en_HK
   - https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/63d3d639-7fd2-4afb-abbe-0d5b5551eef8?redirectedfrom=MSDN
[...]

PR #441, opened 2019-10-25
fix: Make IsMobilePhone supports both string and array locale

liuzhaowei55:
fix: Modify 'en-HK' to 'zh-HK'
feat: IsMobilePhone supports both string and array locale

close issues #440


vlapo:
Why did you close this PR? IsMobilePhone should accept arrays.

liuzhaowei55:
This PR contains controversial commit : https://github.com/typestack/class-validator/issues/440
so I closed this PR

PR #444, opened 2019-10-27
feat: add isJWT validator

Kiliandeca:
Hello, this PR should add the isJWT validator using ValidatorJs
I've updated the doc and added some test tell me if I missed something


vlapo:
Thank you!

PR #445, opened 2019-10-27
feat: add isHash validator

Kiliandeca:
Hello, this PR should add the isHash validator using ValidatorJs
I've updated the doc
I've added tests but could you check if the tests need changes please ?

note: it doesn't support hash with uppercase letters yet, this was added to validatorjs (validatorjs/validator.js#1062) but is not released yet


vlapo:
Looks like new `validator.js@12` is out. I have to check changes but It will be merget into class-validator soon. Thank you for PR

PR #460, opened 2019-11-14
feat: add options for isISO8601 validator

ihorTymofieiev:
This PR is a reference to #412 . Now we can use the option strict = true for additional checks for a valid date, e.g. invalidates dates like 2019-02-29.

### DIFF
--- a/README.md
+++ b/README.md
@@ -836,7 +836,7 @@ validator.isIP(str, version); // Checks if the string is an IP (version 4 or 6).
 validator.isPort(str); // Check if the string is a valid port number.
 validator.isISBN(str, version); // Checks if the string is an ISBN (version 10 or 13).
 validator.isISIN(str); // Checks if the string is an ISIN (stock/security identifier).
-validator.isISO8601(strict, str); // Checks if the string is a valid ISO 8601 date. Use the option strict = true for additional checks for a valid date, e.g. invalidates dates like 2019-02-29.
+validator.isISO8601(str, options); // Checks if the string is a valid ISO 8601 date. Use the option strict = true for additional checks for a valid date, e.g. invalidates dates like 2019-02-29.
 validator.isJSON(str); // Checks if the string is valid JSON (note: uses JSON.parse).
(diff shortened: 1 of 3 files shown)


src/validation/Validator.ts
```diff
@@ -717,9 +717,10 @@ export class Validator {
     /**
      * Checks if the string is a valid ISO 8601 date.
      * If given value is not a string, then it returns false.
+     * Use the option strict = true for additional checks for a valid date, e.g. invalidates dates like 2019-02-29.
      */
-    isISO8601(value: unknown): boolean {
-        return typeof value === "string" && this.validatorJs.isISO8601(value);
+    isISO8601(value: unknown, strict?: boolean): boolean {
```
vlapo: Please use `options?: ValidatorJS.IsISO8601Options` instead of only one `strict` argument. It cover cases when validatorjs add new options in future.

README.md
```diff
@@ -836,7 +836,7 @@ validator.isIP(str, version); // Checks if the string is an IP (version 4 or 6).
 validator.isPort(str); // Check if the string is a valid port number.
 validator.isISBN(str, version); // Checks if the string is an ISBN (version 10 or 13).
 validator.isISIN(str); // Checks if the string is an ISIN (stock/security identifier).
-validator.isISO8601(str); // Checks if the string is a valid ISO 8601 date.
+validator.isISO8601(strict, str); // Checks if the string is a valid ISO 8601 date. Use the option strict = true for additional checks for a valid date, e.g. invalidates dates like 2019-02-29.
```
vlapo: Inputs are in wrong order.

src/validation/Validator.ts
ihorTymofieiev: You're right, I also thought about this in the morning)

vlapo:
Thank you!

jpinkster:
This is awesome! Just ran into this need and there you are! Saving the day. Nice work team 👏

Issue #309, opened 2019-01-24
ValidationError class is missing required "constraints" property.


dylansmurray:
Given the following:

```
import { IsArray, ValidateNested, IsString, IsDefined, validate } from 'class-validator';

class Alpha {
  @IsArray()
  @ValidateNested({ each: true })
  public foo: Array<Beta>;
}

class Beta {
  @IsString()
  @IsDefined()
  public bar : string;
}

const alphaInstance = new Alpha();

const betaInstance1 = new Beta();
betaInstance1.bar = 'valid string';

const betaInstance2 = new Beta();

alphaInstance.foo = [betaInstance1, betaInstance2];

validate(alphaInstance).then((errors) => {
  console.log('Full error:', errors[0])
  console.log('Constraints value:', errors[0].constraints)
});
```

console log result:
```
Full error: ValidationError {
  target: Alpha { foo: [ [Object], Beta {} ] },
  value: [ Beta { bar: 'valid string' }, Beta {} ],
  property: 'foo',
  children:
   [ ValidationError {
       target: [Array],
       value: Beta {},
       property: '1',
       children: [Array] } ] }

Constraints value: undefined
```

`constraints` is defined as a required property -- something like `Object.keys(constraints).forEach` therefore can lead to an unexpected runtime exception.

mbfisher:
+1 for this - just had the same issue!

vlapo:
This works as intended. I think we should mark `constraints` as optional such a higher level of `ValidationError` may not contains constraints.

PR #465, opened 2019-11-18
fix: optional `constraints` property in ValidationError

alburkerk:
## Motivation

I had this log method when an error was appearing : 

```ts
function logErrors(fieldErrors: { requestField: string; error: ValidationError }[]): void {
    console.log(fieldErrors);
    const errorMessages = fieldErrors
        .map(({ requestField, error }) => `- ${requestField}: ${Object.values(error.constraints)}`)
        .join('\n\t');
    logger.error(new Error(`Bad request.\n\t${errorMessages}`));
}
``` 

This triggered a runtime error when error.constraints was undefined. This is the kind of error that could be avoided by Typescript check if the field was optional in ts declaration

## Error reproduction

ValidationError.constraint can be undefined if there is only a nested validation error

## Others tracks of resolution

Maybe instead of changing the declaration, it's better to make sure the field is never undefined. We can discuss about this and I will modify the PR

### DIFF
--- a/src/validation/ValidationError.ts
+++ b/src/validation/ValidationError.ts
@@ -25,7 +25,7 @@ export class ValidationError {
     /**
      * Constraints that failed validation with error messages.
      */
-    constraints: {
+    constraints?: {
         [type: string]: string
     };
 


vlapo:
I agree. Issue already exists https://github.com/typestack/class-validator/issues/309. But this kind of change is a small breaking change. Will merge this to `0.12.0` release.

PR #492, opened 2019-12-19
Change @IsOptional to check that property exists

### LINKED ISSUE #491: @IsOptional with @IsNotEmpty is not working
This code:
```typescript
import { IsNotEmpty, IsOptional, IsString, validate } from 'class-validator';
import { plainToClass } from 'class-transformer';

const PATCH = 'patch';
const POST = 'post';

export class Test {
  @IsOptional({ groups: [PATCH] })
  @IsNotEmpty({ always: true })
  @IsString()
  name: string;
}

async function getValidationErrors(obj, group) {
  return await validate(plainToClass(Test, obj), { groups: [group] });
}

describe('Test', () => {
  it('should fail on post without name', async () => {
    const errors = await getValidationErrors({}, POST);
    expect(errors).not.toEqual([]);
  });
  it('should fail on post when name is undefined', async () => {
    const errors = await getValidationErrors({ name: undefined }, POST);
    expect(errors).not.toEqual([]);
  });
  it('should fail on post when name is null', async () => {
    const errors = await getValidationErrors({ name: null }, POST);
    expect(errors).not.toEqual([]);
  });
  it('should fail on post when name is empty', async () => {
    const errors = await getValidationErrors({ name: '' }, POST);
    expect(errors).not.toEqual([]);
  });
  it('should succeed on patch without name property', async () => {
    const errors = await getValidationErrors({}, PATCH);
    expect(errors).toEqual([]);
  });
  it('should fail on patch when name is undefined', async () => {
    const errors = await getValidationErrors({ name: undefined }, PATCH);
    expect(errors).not.toEqual([]);
  });
  it('should fail on patch when name is null', async () => {
    const errors = await getValidationErrors({ name: null }, PATCH);
    expect(errors).not.toEqual([]);
  });
  it('should fail on patch when name is empty', async () => {
    const errors = await getValidationErrors({ name: '' }, PATCH);
    expect(errors).not.toEqual([]);
  });
});
```
gives the following results:
Screen Shot 2019-12-19 at 18 15 43
Am I missing something here?

Looking at the code of @IsOptional I see that it checks that the property is not undefined or null, that's why the test for empty string is not failing, but in theory shouldn't it check that this property exists on the object?

vsternbach:
closes #491

### DIFF
--- a/src/decorator/decorators.ts
+++ b/src/decorator/decorators.ts
@@ -233,7 +233,7 @@ export function IsOptional(validationOptions?: ValidationOptions) {
             target: object.constructor,
             propertyName: propertyName,
             constraints: [(object: any, value: any) => {
-                return object[propertyName] !== null && object[propertyName] !== undefined;
+                return object.hasOwnProperty(propertyName);
             }],
             validationOptions: validationOptions
         };


NoNameProvided:
Closing as this won't get merged in its current state.

PR #512, opened 2020-01-27
StringEnum support

Samox:
Feature request:
The @IsEnum does not show the possible values during validation, it says "someProperty must be a valid enum value".
When all the values of an enum are strings we could display something similar to the @IsIn decorator.


vlapo:
So target of this PR is better error message (with possible valid values)? We shoul only modify `@IsEnum` message and do not create new validator for this.

vlapo:
Closing for inactivity.

PR #513, opened 2020-01-29
fix: switch isLatitude & isLongitude validators

vschoener:
Hey guys,

I was looking where these functions come from as from `validator.js` they don't exist, but I found out the type between them was inverted.

Tests are ok when changing those. Maybe I can add a few more to be sure it won't happen again?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "class-validator",
-  "version": "0.10.2",
+  "version": "0.11.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


vschoener:
Here is what I got when testing a 400:
`"isLongitude": "latitude must be a longitude string or number"`

crux153:
Any chance to get this merged? It seems to be a simple fix.

johannesschobel:
#537 tries to fix this issue as well.. i think, however, the type assignment (that was fixed in this PR) needs to be changed in all cases ;) this looks like a simple copy/paste error..

vlapo:
@vschoener could you please revert changes in package-lock.json?

vlapo:
Thank you!

Issue #508, opened 2020-01-21
IsNumberString ignores maxDecimalPlaces option


y-moroz:
`IsNumberString` decorator  accepts `IsNumberOptions` as the second argument but ignores maxDecimalPlaces option. 
So if we have next class:
```typescript
export class Dto {
  @IsNumberString({}, {maxDecimalPlaces: 1})
  id: string;
}
```
And next object: 
```typescript
{
  id: "1.123"
}
```
ER: This object is invalid because there are 3 decimal digits.
AR: The object is valid.

vlapo:
Related PR https://github.com/typestack/class-validator/pull/518

vlapo:
`IsNumberString` accepts `IsNumericOptions` not `IsNumberOptions`. Fixed by https://github.com/typestack/class-validator/commit/62b993f3d360c8844687c8856ae9d84bb50b8654

PR #518, opened 2020-02-04
Fix issue on IsNumberString ignores maxDecimalPlaces option #508

Getriax:
**Now**
`IsNumberString` accepts `IsNumberOptions` options but is NOT passing and validating against the number validator.

**Fix**
Added the support for a number validation with `IsNumberString`. The value will be also validated against provided `IsNumberOptions` as a float.

### DIFF
--- a/src/validation/Validator.ts
+++ b/src/validation/Validator.ts
@@ -181,7 +181,7 @@ export class Validator {
             case ValidationTypes.IS_BOOLEAN_STRING:
                 return this.isBooleanString(value);
             case ValidationTypes.IS_NUMBER_STRING:
-                return this.isNumberString(value);
+                return this.isNumberString(value, metadata.constraints[0]);
 
             /* string checkers */
             case ValidationTypes.CONTAINS:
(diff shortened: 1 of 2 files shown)


test/functional/validation-functions-and-decorators.spec.ts
```diff
@@ -3281,19 +3294,19 @@ describe("isHash", function() {
         it("should not fail if validator.validate said that its valid", function(done) {
             checkValidValues(new MyClass(), validValues, done);
         });
-    
+
```
vlapo: please remove this "empty" changes

src/validation/Validator.ts
```diff
@@ -530,8 +530,8 @@ export class Validator {
      * Checks if the string is numeric.
      * If given value is not a string, then it returns false.
      */
-    isNumberString(value: unknown, options?: ValidatorJS.IsNumericOptions): boolean {
-        return typeof value === "string" && this.validatorJs.isNumeric(value, options);
+    isNumberString(value: unknown, options?: IsNumberOptions): boolean {
+        return typeof value === "string" && this.validatorJs.isNumeric(value) && (options == null || this.isNumber(parseFloat(value), options));
```
vlapo: It is necessary ` && (options == null || this.isNumber(parseFloat(value), options)` part? I think `isNumber` accept also undefined as option argument.

src/validation/Validator.ts
```diff
@@ -530,8 +530,8 @@ export class Validator {
      * Checks if the string is numeric.
      * If given value is not a string, then it returns false.
      */
-    isNumberString(value: unknown, options?: ValidatorJS.IsNumericOptions): boolean {
-        return typeof value === "string" && this.validatorJs.isNumeric(value, options);
+    isNumberString(value: unknown, options?: IsNumberOptions): boolean {
```
vlapo: We use `ValidatorJS.*` prefix everywhere. Please do not change it her.

PR #524, opened 2020-02-12
fix: IsNumber validator now works when maxDecimalPlaces=0

### LINKED ISSUE #523: IsNumber cannot check for zero decimal places
checking for zero decimal places does not work:

```typescript
    class ZeroDecimalPlacesTest {
        @IsNumber({ maxDecimalPlaces: 0 })
        someProperty: number;
    }
```

The following validation logic is wrong: 
https://github.com/typestack/class-validator/blob/90a663866eaeb5a0cb38cce5501e0aa28984ed34/src/validation/Validator.ts#L438

instead it  should check if `maxDecimalPlaces` is `undefined`:
```typescript
if (options.maxDecimalPlaces !== undefined) {
```

tmtron:


vlapo:
Please update your PR.
 - update only code related to this PR
 - do not update package.json
 - do not update indents

tmtron:
I've undone changes to `package.json`

I did not change the indentation - it seems my IDE automatically removes trailing whitespace (on save/commit) - I could not find a way to switch this off. You should conisder using [prettier](https://prettier.io/) to get consistent formatting.

Issue #331, opened 2019-03-14
Proposal: Negative @Matches() validator


mazyvan:
I have to change a regex just to match the inverse of it. It would be great if by default there was something like `@NotMatches()` for instance.

I could open a PR this Saturday if you like the idea.

Anyway. Great project. Keep it up 👍

vlapo:
I dont think we need `NotMatches` validator. It is simple write regex as negative version of another regex.

PR #527, opened 2020-02-13
feat: add NotMatches validator

Kiliandeca:
Hello, this should add a NotMatches decorator as suggested in issue #331 

* [x] Documentation (Validator and Manual validation)
* [x] Validator and Decorator
* [x] Default validation error message
* [x] Tests

Have a good day


vlapo:
Close as this is only derivation of `Matches` decorator and it is simple write negative version of regex.
Thank you for your contribution.

Issue #528, opened 2020-02-14
question: how skip missing properties on nested objects?


CulmiDoom:
If i have a structure like this:
```typescript
class Foo {
  @IsNumber()
  a: number
  
  @ValidateNested()
  @Type(() => Bar)
  bar: Partial<Bar>
}

class Bar {
  @IsNumber()
  c: number
  @IsNumber()
  d: number
}
```

how do I make `validate()` call so Foo gets fully validated, but for nested `.bar` property it skips missing fields? 

(i.e. if `foo.a` is missing it will throw an error, but if `foo.bar.c` is missing it will not)

NoNameProvided:
You can use the `@IsOptional` decorator. Minimal working example.

```ts
import 'reflect-metadata';
import { validateOrReject, IsNumber, ValidateNested, IsOptional } from 'class-validator';

class ExampleA {

  @IsNumber()
  outerProperty: number
  
  @ValidateNested()
  bar: Partial<ExampleB>
}

class ExampleB {

  @IsOptional()
  @IsNumber()
  childPropertyA?: number;

  @IsOptional()
  @IsNumber()
  childPropertyB?: number;
}

const instance = new ExampleA();

instance.outerProperty = 5;
instance.bar = new ExampleB();
instance.bar.childPropertyA = 5;

validateOrReject(instance)
  .then(console.log)
  .catch(console.error)
// validation passes as expected because childPropertyB is not required
// if we would change the type of one of the existing properties it would throw an error as expected
```

NoNameProvided:
Closing this as solved. If your issue still persists, please comment on this issue again and I will reopen it.

teidesu:
but what if i want to re-use `ExampleB` in other place but require all fields to be present?

NoNameProvided:
We have groups for that scenario. You can find out more about them via reading the docs in the readme.

PR #529, opened 2020-02-15
feat: add partials support for nested

teidesu:
so basically implemented #528 --  nesteds can now be validated as partials:
```typescript
export class Post {
    @ValidateNested({ partial: true })
    content: Partial<Content>

    @IsNumber()
    ownerId: number
}
export class Content {
    @IsString()
    @MaxLengh(1000)
    text: string
    
    @IsUrl()
    image: string
}

const post = new Post()
post.ownerId = 42
post.content = new Content()
post.content.text = 'ok'

validate(post) // ok
```

not quite sure whether what i have done is the best way of achieving this, but at least it seems to work.


NoNameProvided:
Closing as wontfix because we can use the `@IsOptional` decorator to achieve the same result. You can see a minimal example in the linked issue.

Issue #502, opened 2020-01-05
Decorators for IsLatitude and IsLongitude have wrong types


scottcollins:
The IsLatitude decorator defines a type of IS_LONGITUDE and the IsLongitude decorator defines a type of IS_LONGITUDE.  Should be the other way around

scottcollins:
I have a pull request ready for this, but I do not have access to push to a branch. how can I get access?

vlapo:
You have to fork this repository, commit in forked one and create PR https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork.

vlapo:
@rubiin could you please check this one? Check if it is real bug and create PR - looks like a simple fix.

rubiin:
@vlapo fixed in pr https://github.com/typestack/class-validator/pull/563

PR #537, opened 2020-03-01
fix: switch isLatitude & isLongitude validators

zesani:
Validatorjs check if the string is a valid latitude-longitude coordinate in the format lat,long or lat, long.

Ref: https://github.com/validatorjs/validator.js

### DIFF
--- a/src/validation/Validator.ts
+++ b/src/validation/Validator.ts
@@ -371,14 +371,14 @@ export class Validator {
     * Checks if a given value is a latitude.
     */
     isLatitude(value: unknown): boolean {
-        return (typeof value === "number" || this.isString(value)) && this.isLatLong(`0,${value}`);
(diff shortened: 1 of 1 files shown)


johannesschobel:
this may be fixed if #513 will be implemented? can you check please? @zesani

vlapo:
https://github.com/typestack/class-validator/issues/502 looks like we have wrong type, too:
```ts
export function IsLatitude(validationOptions?: ValidationOptions): PropertyDecorator {
    return function (object: Object, propertyName: string) {
        const args: ValidationMetadataArgs = {
            type: ValidationTypes.IS_LONGITUDE,
            target: object.constructor,
            propertyName: propertyName,
            validationOptions: validationOptions
        };
        getMetadataStorage().addValidationMetadata(new ValidationMetadata(args));
    };
}
```
@zesani could you also fix that and update tests? We should cover edge cases.

vlapo:
Thank you!

Issue #538, opened 2020-03-01
Validate array of arrays


Eldar-X:
Hello, I am trying to validate the arrays in arrays but I could not find a correct method.

The structure looks like this:

```typescript
class MyInnerDto {
 @IsString()
  name: string;
}
 
class TestDto {
@IsArray({ each: true })
 myArray:MyInnerDto[][];
}
```

jBouyoud:
Hello,

I've the same kind of issue with Map and Array values. 
The test case below show validation error on `nestedMap` but not on `nestedWithArray`.
Is there another tips to validate this kind of structure ? or it's a bug ?

With `class-validator": "0.11.0"`

```typescript
import { IsString, ValidateNested, Validator } from 'class-validator';

describe('NestedValidation', () => {
  class NestedClass {
    @IsString()
    name: string;

    constructor(name: any) {
      this.name = name;
    }
  }

  class MasterClass {
    @ValidateNested({ each: true })
    nestedMap: Map<string, NestedClass>;

    @ValidateNested({ each: true })
    nestedWithArray: Map<string, NestedClass[]>;

    constructor(
      nestedMap: Map<string, NestedClass>,
      nestedWithArray: Map<string, NestedClass[]>,
    ) {
      this.nestedMap = nestedMap;
      this.nestedWithArray = nestedWithArray;
    }
  }

  describe('validation', () => {
    let validator: Validator;

    beforeAll(() => {
      validator = new Validator();
    });

    it('should validate nested object array', async () => {
      const nestedMap = new Map<string, NestedClass>();
      nestedMap.set('test', new NestedClass(42));
      const nestedMapWithArray = new Map<string, NestedClass[]>();
      nestedMapWithArray.set('test', [new NestedClass(42)]);
      const masterClass = new MasterClass(nestedMap, nestedMapWithArray);

      const errors = await validator.validate(masterClass);

      function expectNestedError(property: string) {
        const nestedError = errors.find(error => error.property === property);
        expect(nestedError).toBeDefined();
        expect(nestedError).toMatchObject({ property, target: masterClass, });
      }
      expectNestedError('nestedMap');
      expectNestedError('nestedWithArray');
    });
  });
});

```

Thanks

vlapo:
https://github.com/typestack/class-validator/pull/539 fixed multi-dimensional arrays for `@ValidatedNested`. But we should also support multi-dimensional arrays for cases like:
```ts
@IsNumber({}, {
	each: true
})
tags: number[][];
[...]

PR #539, opened 2020-03-03
fix: ValidateNested support multi-dimensional arrays

jBouyoud:
@NoNameProvided This PR aims to fix https://github.com/typestack/class-validator/issues/538

### DIFF
--- a/test/functional/nested-validation.spec.ts
+++ b/test/functional/nested-validation.spec.ts
@@ -2,7 +2,6 @@ import "es6-shim";
 import {Contains, IsDefined, MinLength, ValidateNested} from "../../src/decorator/decorators";
 import {Validator} from "../../src/validation/Validator";
 import {expect} from "chai";
-import {inspect} from "util";
 import {ValidationTypes} from "../../src/validation/ValidationTypes";
 
 import {should, use } from "chai";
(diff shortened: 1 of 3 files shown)


jBouyoud:
@vlapo

test/functional/nested-validation.spec.ts
```diff
@@ -67,6 +66,12 @@ describe("nested validation", function () {
 
             @ValidateNested()
             mySubClasses: MySubClass[];
+
+            @ValidateNested()
+            mySubSubClasses: MySubClass[][];
+
+            @ValidateNested()
+            mySubSubSubClasses: MySubClass[][][];
```
satanTime: does it work with tuples too? `myTupleClasses: [MyClass1, MyClass2];`

vlapo: `myTupleClasses: [MyClass1, MyClass2];` is basically simple array (we already support simple arrays). But we validate only items of array. So for example this will pass:
```ts
class Tuple1 {
  @IsNumber()
  num: number;
}

class Tuple2 {
  @IsString()
  name: string;
}

class Test {

  @ValidateNested()
  tuples: [Tuple1, Tuple2];

}

const t = new Test();
t.tuples = [new Tuple1(), new Tuple1() as any];
t.tuples[0].num = 2 as any;
(t.tuples[1] as any).num = 2 as any;

console.log(validateSync(t));
```

jBouyoud:
I've added a very simple example about multi dimensional array. 
Let me know if that not enough

vlapo:
This is not full support of multi-dimensional arrays but at least @ValidateNested works now.
So I will not close #538 and track multi-dimensional support there.
Thank you!

PR #548, opened 2020-03-09
feat: add IsFirebasePushId validator

elmarti:
This PR adds support for Firebase-style PushIds, you can see more about the science and inner-workings here: https://firebase.googleblog.com/2015/02/the-2120-ways-to-ensure-unique_68.html

**Overview**
Simply, we're checking:
- The input must be a string
- The string must be 20 characters in length
- The string must only contain web-safe characters

I've checked the backlog and PRs for duplicates and found nothing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -854,7 +854,7 @@ validator.isMultibyte(str); // Checks if the string contains one or more multiby
 validator.isSurrogatePair(str); // Checks if the string contains any surrogate pairs chars.
 validator.isURL(str, options); // Checks if the string is an url.
 validator.isUUID(str, version); // Checks if the string is a UUID (version 3, 4, 5 or all).
-validator.isPushId(str); // Checks if the string is a Push Id
+validator.IsFirebasePushId(str); // Checks if the string is a Firebase Push Id
 validator.isUppercase(str); // Checks if the string is uppercase.
 validator.length(str, min, max); // Checks if the string's length falls in a range.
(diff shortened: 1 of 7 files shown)


test/functional/reject-validation.spec.ts
```diff
@@ -1,4 +1,5 @@
 import "es6-shim";
+import "es6-shim";
```
vlapo: duplicate row

README.md
```diff
@@ -951,6 +952,7 @@ validator.isInstance(value, target); // Checks value is an instance of the targe
 | `@IsSurrogatePair()`                            | Checks if the string contains any surrogate pairs chars.                                                                         |
 | `@IsUrl(options?: IsURLOptions)`                | Checks if the string is an url.                                                                                                  |
 | `@IsUUID(version?: "3"\|"4"\|"5"\|"all")`              | Checks if the string is a UUID (version 3, 4, 5 or all ).                                                                              |
+| `@IsPushId()`                                   | Checks if the string is a Firebase Push id                                                                                      |
```
vlapo: add link to firebase PushId specification

vlapo:
Also name `IsPushId` is really misleading. What about `IsFirebasePushId` ?

elmarti:
@vlapo Agreed, although the Firebase part was excluded as they can exist and be used apart from Firebase, however in practice they're not really.

PR #564, opened 2020-03-26
fix: accept negative timezone in isDateString

alencarandre:
# Issue
https://github.com/typestack/class-validator/issues/565

# Expected behavior

String like "2018-01-04T08:15:30-04:00" gives valid

# Actual behavior

String like "2018-01-04T08:15:30-04:00" gives invalid

# Solution

This PR add `-` sign to regex in order to validation accept negative timezones


vlapo:
Thank you!

Issue #281, opened 2018-11-09
Decorator ValidateNested is not working


henrikra:
I am using `class-validator` version `0.9.1`.

Here I have simple case where I have `RootObject` which has property `myEpicNestedObject` and we want to validate that nested object. Then [according to documentation](https://github.com/typestack/class-validator#validating-nested-objects) you should use `ValidateNested` but isn't actually doing anything.

Here is a simple sample code to try this out:
```ts
import { IsInt, ValidateNested, validate, IsString } from 'class-validator';

class NestedObject {
  @IsInt()
  myCoolNumber!: number;
}

class RootObject {
  @IsString()
  myAwesomeName!: string;

  @ValidateNested()
  myEpicNestedObject!: NestedObject;
}

const someObject = new RootObject();
someObject.myAwesomeName = 'lol';
someObject.myEpicNestedObject = { myCoolNumber: 1.2 }; // should fail because not integer

validate(someObject)
  .then(errors => {
    console.log('all fine', errors);
  })
  .catch(error => {
    console.log(error);
  });

```

What is wrong with `ValidateNested` or am I misusing it?

19majkel94:
nested object has to be an instance, created using `new`

skrosoft:
Yes, it works using `new `... I could be great if this could works without the `new` operator ...

henrikra:
@19majkel94 You were correct!

PR #571, opened 2020-03-31
docs: added note about nested validation and object creation

dooman87:
I think it would be handy to have the limitation described in this PR. Took me almost an hour to find closed issue #281.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,10 +313,12 @@ import {ValidateNested} from "class-validator";
 export class Post {
 
     @ValidateNested()
-    user: User;
+    user: User = new User();
 
 }
 ```
+
+Please note that nested object **must** be created using `new` keyword, otherwise `@ValidateNested` won't work.
 
 It also works with multi-dimensional array, like :
 
--- a/README.md
+++ b/README.md
@@ -313,12 +313,12 @@ import {ValidateNested} from "class-validator";
 export class Post {
 
     @ValidateNested()
-    user: User = new User();
+    user: User;
 
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -313,11 +313,13 @@ import {ValidateNested} from "class-validator";
 export class Post {
 
     @ValidateNested()
-    user: User;
+    user: User = new User();
 
 }
 ```
 
+Please note that nested object **must** be created using `new` keyword, otherwise `@ValidateNested` won't work.
```
vlapo, line 321: I think this note is a little bit misleading. `new` keyword is not the only way how to create instance of class (e.g. `class-transformer`). `class-validator` use `object.constructor` as target for validation so we should change this note to something like:

Please note that nested object **must** be instance of class, otherwise `@ValidateNested` won't know what class is target of validation. Check also [Validating plain objects](https://github.com/typestack/class-validator#validating-plain-objects).

Or something like that.

README.md
```diff
@@ -313,11 +313,13 @@ import {ValidateNested} from "class-validator";
 export class Post {
 
     @ValidateNested()
-    user: User;
+    user: User = new User();
```
vlapo, line 316: Again misleading code. I think note below this section is enough.

dooman87: Removed.

README.md
dooman87: Thanks @vlapo for reviewing. Your option sounds much better. I update PR, please have a look

vlapo:
Thank you!

Issue #278, opened 2018-11-04
fix: `@ValidateIf` receives transformed nested object instead of POJO


tsamsiyu:
I have the following class definitions:

```
class Field {
    @IsString()
    readonly name: string;

    @IsString()
    @IsIn(['values', 'numbers', 'dates'])
    readonly type: string;

    @ValidateNested()
    @Type(() => Rule)
    @IsNotEmpty()
    readonly rules: Rule[];
}

class Rule {
    @ValidateIf((a, b) => { console.log(b); return true; })
    @IsDate()
    readonly fromDate: Date;

    @IsDate()
    readonly toDate: Date;

    @IsNumber()
    readonly fromNumber: number;

    @IsNumber()
    readonly toNumber: number;

    @IsString()
    readonly label: string;

    @IsNumber()
    readonly value: number;

    @IsNumber()
    readonly rate: number;
}

```
Validating properties of class `Rule` directly depends on value of property `type` of `Field` class object but in callback of `ValidateIf` decorator I've got object of `Rule` class. So it makes impossible to achieve conditional validation on nested objects.

ChrisGatzo:
Anyone on this?

hlibco:
+1 Anyone on this?

hlibco:
@NoNameProvided I'm considering to submit a PR for this issue, is that okay? I'd need some guidance tho. I've explored the internals and an outlined solution can be:

1. Add a `context` to the signature of the `ValidateIf` decorator (`decorators.ts`).
```
export function ValidateIf(condition: (object: any, value: any, context: any) => boolean, validationOptions?: ValidationOptions)
```

2. Pass the context to the validator in the `conditionalValidations()` (`ValidationExecutor.ts`):
```
.map(metadata => metadata.constraints[0](object, value, metadata.context))
```

In the use-case given by @tsamsiyu, the ValidateIf decorator should be used like following, in case we want to validate `fromDate` ONLY IF the property `name` in the Field class is provided:

```
class Rule {
    @ValidateIf((obj, value, context) => { !!return context.Field.name })
    @IsDate()
    readonly fromDate: Date;
}
```
Does it look the right way to approach this problem?

hlibco:
After digging a bit more in the internals, I realized that the approach I mentioned above would not work.

@NoNameProvided I submitted a PR with tests included. I'd appreciate if you can check it out and give your feedback. Btw, congrats for the project, I'm a big fan of it.

https://github.com/typestack/class-validator/pull/301

[...]

Issue #582, opened 2020-04-18
Conditional validation


satanTime:
Below you can find an example of the issue. I always want both fields to be of the boolean type and always require at least one of them to be `true`.
I couldn't find a way how to do it. Please let me know if it exists.

My suggestion is to implement `ValidateWithGroups(condition: (object: any, value: any) => string | string[] | undefined, validationOptions?: ValidationOptions)` decorator that accepts a callback that returns `group` or `groups` that should be additionally verified on the current field.

A test to show the issue and a possible solution:
```typescript
describe("validation with groups", function() {
    it("should respect another side", function() {
        class MyClass {
            @IsBoolean()
            @IsNotEmpty()
            @ValidateIf((object: MyClass) => !object.isDisabled)
            isEnabled: boolean;

            @IsBoolean()
            @IsNotEmpty()
            @ValidateIf((object: MyClass) => !object.isEnabled)
            isDisabled: boolean;
        }

        const model = new MyClass();
        model.isEnabled = true;
        model.isDisabled = <any>"impossible";
        return validator.validate(model).then(errors => {
            errors.length.should.be.equal(1); // but it is 0
        });
    });

    it("should have a solution", function() {
        class MyClass {
            @IsBoolean()
            @IsNotEmpty({
                groups: ["extra"],
            })
            // returns groups for validation or undefined for the default behavior.
            @ValidateWithGroups((object: MyClass, value: boolean) => object.isDisabled ? undefined : ["extra"])
            isEnabled: boolean;

            @IsBoolean()
            @IsNotEmpty({
                groups: ["extra"],
            })
            // returns groups for validation or undefined for the default behavior.
            @ValidateWithGroups((object: MyClass, value: boolean) => object.isEnabled ? undefined : ["extra"])
            isDisabled: boolean;
        }

        const model = new MyClass();
        model.isEnabled = true;
        model.isDisabled = <any>"impossible";
        return validator.validate(model).then(errors => {
            // [
            //   ValidationError {
            //     target: MyClass { isEnabled: true, isDisabled: 'impossible' },
            //     value: 'impossible',
            //     property: 'isDisabled',
            //     children: [],
            //     constraints: { isBoolean: 'isDisabled must be a boolean value' }
            //   }
            // ]
            errors.length.should.be.equal(0);
        });
    });
});
```

@vlapo, what do you think? If it's fine for you I can start to work on the implementation.

braaar:
implemented in https://github.com/typestack/class-validator/pull/1579

PR #584, opened 2020-04-19
feat: add support for conditional nested validation

davidthor:
I have a configuration syntax that my users use to configure our platform, and was frequently running into cases where I wanted to conditionally validate a nested object (we frequently support a shorthand syntax for `<property>.default = <default-value>` as `<property> = <default-value>` when the user doesn't intend to use the rest of the nested config block). The suggestion to use "groups" from https://github.com/typestack/class-validator/issues/278 seemed insufficient since it requires me to explicitly choose which groups to validate against when really I just want the users input to guide the validation. The changes here would allow for a new validator, `@ValidateNestedIf()` that would only trigger nested validation upon the passing of the embedded condition.

### DIFF
--- a/test/functional/nested-validation.spec.ts
+++ b/test/functional/nested-validation.spec.ts
@@ -329,10 +329,7 @@ describe("nested validation", function () {
         }
 
         class MyClass {
-            @ValidateNestedIf((_, value: any) => {
-                console.log(value);
-                return value instanceof MySubClass;
-            })
+            @ValidateNestedIf((_, value: any) => value instanceof MySubClass)
             mySubClass: String | MySubClass;
         }
 
--- a/test/functional/nested-validation.spec.ts
+++ b/test/functional/nested-validation.spec.ts
@@ -1,12 +1,11 @@
-import "es6-shim";
-import {Contains, IsDefined, MinLength, ValidateNested} from "../../src/decorator/decorators";
-import {Validator} from "../../src/validation/Validator";
-import {expect} from "chai";
-import {ValidationTypes} from "../../src/validation/ValidationTypes";
-
-import {should, use } from "chai";
-
+import { expect, should, use } from "chai";
 import * as chaiAsPromised from "chai-as-promised";
+import "es6-shim";
+import { Contains, IsDefined, MinLength, ValidateNested, ValidateNestedIf } from "../../src/decorator/decorators";
+import { ValidationTypes } from "../../src/validation/ValidationTypes";
+import { Validator } from "../../src/validation/Validator";
+
+
 
 should();
 use(chaiAsPromised);
@@ -321,6 +320,46 @@ describe("nested validation", function () {
             subSubError.value.should.be.equal("my");
         });
 
+    });
+
+    it("should optionally validate nested options", () => {
+        class MySubClass {
+            @MinLength(5)
+            name: string;
+        }
+
+        class MyClass {
+            @ValidateNestedIf((_, value: any) => {
+                console.log(value);
+                return value instanceof MySubClass;
+            })
+            mySubClass: String | MySubClass;
+        }
+
+        const model = new MyClass();
+        model.mySubClass = new MySubClass();
+        model.mySubClass.name = "my";
+
+        const model2 = new MyClass();
+        model2.mySubClass = "my";
+
+        return validator.validate(model).then(errors => {
+            errors.length.should.be.equal(1);
+
+            errors[0].target.should.be.equal(model);
+            errors[0].property.should.be.equal("mySubClass");
+            errors[0].value.should.be.equal(model.mySubClass);
+            expect(errors[0].constraints).to.be.undefined;
+            const subError1 = errors[0].children[0];
+            subError1.target.should.be.equal(model.mySubClass);
+            subError1.property.should.be.equal("name");
+            subError1.constraints.should.be.eql({minLength: "name must be longer than or equal to 5 characters"});
+            subError1.value.should.be.equal("my");
+
+            return validator.validate(model2).then(errors => {
+                errors.length.should.be.equal(0);
+            });
+        });
     });
 
 });
--- a/README.md
+++ b/README.md
@@ -333,6 +333,23 @@ export class Plan2D {
 }
 ```
 
+If your property only requires nested validation in some cases (e.g. the property can be one of several different types), you can conditionally validate nested:
+
+```typescript
+import {ValidateNestedIf, MinLength} from "class-validator";
+
+export class User {
+    @MinLength(5)
+    name: String;
+}
+
+export class Post {
+
+    @ValidateNestedIf((object, value) => value instanceof User)
+    user: String | User;
+}
+```
+
 ## Validating promises
 
 If your object contains property with `Promise`-returned value that should be validated, then you need to use the `@ValidatePromise()` decorator:
--- a/src/decorator/common/ValidateNested.ts
+++ b/src/decorator/common/ValidateNested.ts
@@ -1,8 +1,8 @@
-import { ValidationOptions } from "../ValidationOptions";
+import { getMetadataStorage } from "../../metadata/MetadataStorage";
+import { ValidationMetadata } from "../../metadata/ValidationMetadata";
 import { ValidationMetadataArgs } from "../../metadata/ValidationMetadataArgs";
 import { ValidationTypes } from "../../validation/ValidationTypes";
-import { ValidationMetadata } from "../../metadata/ValidationMetadata";
-import { getMetadataStorage } from "../../metadata/MetadataStorage";
+import { ValidationOptions } from "../ValidationOptions";
 
 /**
  * Objects / object arrays marked with this decorator will also be validated.
@@ -17,6 +17,23 @@ export function ValidateNested(validationOptions?: ValidationOptions): PropertyD
             type: ValidationTypes.NESTED_VALIDATION,
             target: object.constructor,
             propertyName: propertyName,
+            validationOptions: opts,
+        };
+        getMetadataStorage().addValidationMetadata(new ValidationMetadata(args));
+    };
+}
+
+export function ValidateNestedIf(condition: (object: any, value: any) => boolean, validationOptions?: ValidationOptions): PropertyDecorator {
+    const opts: ValidationOptions = { ...validationOptions };
+    const eachPrefix = opts.each ? "each value in " : "";
+    opts.message = opts.message || eachPrefix + "nested property $property must be either object or array";
+
+    return function (object: Object, propertyName: string) {
+        const args: ValidationMetadataArgs = {
+            type: ValidationTypes.NESTED_VALIDATION,
+            target: object.constructor,
+            propertyName: propertyName,
+            constraints: [condition],
             validationOptions: opts,
         };
         getMetadataStorage().addValidationMetadata(new ValidationMetadata(args));
--- a/src/register-decorator.ts
+++ b/src/register-decorator.ts
@@ -1,12 +1,12 @@
-import {ConstraintMetadata} from "./metadata/ConstraintMetadata";
-import {ValidatorConstraintInterface} from "./validation/ValidatorConstraintInterface";
-import {ValidationMetadata} from "./metadata/ValidationMetadata";
-import {ValidationMetadataArgs} from "./metadata/ValidationMetadataArgs";
-import {ValidationTypes} from "./validation/ValidationTypes";
-import {ValidationArguments} from "./validation/ValidationArguments";
 import { getFromContainer } from "./container";
-import { MetadataStorage, getMetadataStorage } from "./metadata/MetadataStorage";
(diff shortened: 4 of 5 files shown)


vlapo:
Hi, I think you could use already exist `ValidateIf` decorator for your case. I updated test case from this PR:

```ts
class MySubClass {
    @MinLength(5)
    name: string;
}

class MyClass {
    @ValidateIf(o => o.mySubClass instanceof MySubClass)
    @ValidateNested()
    mySubClass: String | MySubClass;
}

const model = new MyClass();
model.mySubClass = new MySubClass();
model.mySubClass.name = "my";

const model2 = new MyClass();
model2.mySubClass = "my";

console.log('model', validateSync(model));
console.log('model2', validateSync(model2));
```

davidthor:
Unfortunately that disables all validation on the field - looks like I didn't capture the use-case with the test case, but I have another custom decorator on my parent classes that looks something like this:

```typescript
class MyClass {
    @ValidateNestedIf((o, v) => v instanceof MySubClass)
    @MatchesOneOf([
        value => typeof value === 'string',
        value => value instanceof MySubClass,
    ])
    mySubClass: String | MySubClass;
}
```

This ensures I can still do checks on the data even if it's not of the nested type.

vlapo:
> This ensures I can still do checks on the data even if it's not of the nested type. In a perfect world, `conditions` would be a part of `ValidationOptions` so I can specify individual checks to run if the data is shaped different ways.

We all want perfect world :) I like an idea about `conditions` property as part of `ValidationOptions`. There are plenty of issues/questions about conditional validation and I think this solution could solve at least some of them.

@satanTime what do you think?

test/functional/nested-validation.spec.ts
```diff
@@ -323,4 +322,44 @@ describe("nested validation", function () {
...
 
+    it("should optionally validate nested options", () => {
+        class MySubClass {
+            @MinLength(5)
+            name: string;
+        }
+
+        class MyClass {
+            @ValidateNestedIf((_, value: any) => {
+                console.log(value);
```
satanTime, line 333: can be omitted from the PR

satanTime:
Hi @vlapo, for the nested objects it looks good. But I can't find `MatchesOneOf` declaration in the master branch.

In the same time, I guess, it doesn't solve the issue of validation of the `string` case, because it can be anything: `number`, `boolean` etc. I'm not sure whether it's possible to check currently.

```typescript
class MyClass {
    @ValidateNestedIf((o, v) => v instanceof MySubClass)
    @MatchesOneOf([
        value => typeof value === 'string',
        value => value instanceof MySubClass,
    ])
    @IsString() // <- would it work?
    mySubClass: String | MySubClass;
}
```

I've opened an issue with a suggestion how we could solve it (https://github.com/typestack/class-validator/issues/582). Then it would be like that.
```typescript
class MyClass {
    @WithGroups((object, value) => typeof value === 'string' ? 'string' : 'object')
    @ValidateNested({groups: ['object']})
    @IsString({groups: ['string']})
    mySubClass: String | MySubClass;
}
```

vlapo:
I added you because I think this is similar problem as you describe in proposal https://github.com/typestack/class-validator/issues/582. In your proposal you want use already existing `groups` feature and extend it with `@WithGroup`. But IMHO groups are not very clear concept for our users. 

What @davidthor proposed is some kind of decorator level conditions. So example should looks like:
```ts
class MyClass {
    @ValidateNested({
        confition: (object, value) => value instanceof MySubClass
    })
    @IsString({
        confition: (object, value) => typeof value === 'string'
    })
    @MatchesOneOf([
        value => typeof value === 'string',
        value => value instanceof MySubClass,
    ])
    mySubClass: String | MySubClass;
}
```

`MatchesOneOf` is custom validation :)

vlapo:
Also adding global validation context we could also remove groups functionality completly. E.g.
```ts
class MyClass {
    @ValidateNested({
        confition: (object, value, context) => {
			return value instanceof MySubClass && context.requestType === 'PUT';
		}
    })
    @IsString({
        confition: (object, value, context) => typeof value === 'string'
    })
    @MatchesOneOf([
        value => typeof value === 'string',
        value => value instanceof MySubClass,
    ])
    mySubClass: String | MySubClass;
}

validate(t, {context: {requestType: 'PUT'}})
```

satanTime:
It definitely makes sense with the `condition`.

Despite that I see 2 minuses:
* we need to copy the callback for every decorator if we want to apply more than 1 rule for a specified type / condition, that means more coding / copy-pasting.
* It will be quite easy to forget the `else` case (the MatchesOneOf) to avoid other types / values, because people will focus on types they need like `string` and `MySubClass` and don't consider what can happen if user sends another value because by default the class-validator works in the way that it doesn't allow anything that wasn't specified with decorators.

vlapo:
> * we need to copy the callback for every decorator if we want to apply more than 1 rule for a specified type / condition, that means more coding / copy-pasting.

This should be resolved by using helper functions:
```ts
function isMySubClass(object: any, value: any) => value instanceof MySubClass;
function isString(object: any, value: any) => typeof value === 'string';
 
class Test {
	@ValidateNested({ condition: isMySubClass })
    @IsString({ condition: isString })
	prop1: string | SubClass1;

	@ValidateNested({ condition: isMySubClass })
    @IsString({ condition: isString })
	prop2: string | SubClass1;
}
```

> * It will be quite easy to forget the `else` case (the MatchesOneOf) to avoid other types / values, because people will focus on types they need like `string` and `MySubClass` and don't consider what can happen if user sends another value because by default the class-validator works in the way that it doesn't allow anything that wasn't specified with decorators.

Sure but same problem is with groups. And it should be a basic skill of developer to handle else cases :)

satanTime:
about the basic skill - true :) I would use the same for the groups :) 

I need to think about the idea with `condition`. It's good, but I'll try to think about how to solve copy paste of helper functions.
Personally I would prefer a single place (a condition resolver decorator) where all conditions are defined for a single field without coping anything and the group thing fits it good.

I'll think over it next days and drop here a message.

davidthor:
@satanTime the `MatchesOneOf()` decorator was just a custom one I have in my own repo. I was looking to keep code introductions to a minimum, and since the mechanics already existed for me to create that decorator without code changes I figured it'd be up to someone else to suggest whether it should be one of the core options.

> ```ts
> function isMySubClass(object: any, value: any) => value instanceof MySubClass;
> function isString(object: any, value: any) => typeof value === 'string';
>  
> class Test {
> 	@ValidateNested({ condition: isMySubClass })
>     @IsString({ condition: isString })
> 	prop1: string | SubClass1;
> 
> 	@ValidateNested({ condition: isMySubClass })
>     @IsString({ condition: isString })
> 	prop2: string | SubClass1;
> }
> ```

@vlapo support for conditions like your example would be perfect! I actually simplified my case and in fact have 3 different types supported (String, Number, and MyCustomClass), and generalizing the conditions under which a validation rule applies seems like a great solution.

As you also said above, I found the `groups` usage to be very confusing – particularly because in my case the desired "group" isn't something I would pass into the `validate()` method, but would instead be something derived from the shape of the config provided by a user. Using groups to accomplish this would require me to parse and determine the input shape _before_ validation which minimizes the value of the validation which provides me the luxury of assuring that the shape is one that is supported before proceeding.

satanTime:
@davidthor, you have been caught in the trap I had meant :) in your code code if the prop1 or prop2 is number, boolean etc it won't fail the validation :)

The idea is good, I like it too, we need to develop it further, I think there is a way how to solve this case, I hope after a few days of brainstorming something pop up in our minds. I have some ideas, but I'll share them later next week.

davidthor:
Ahhh, I totally see it now! Whatever concept of "conditional" that exists also needs a concept of "default" behavior to fallback to, otherwise we'd have no way of ever rejecting non-matching input. I'll put some more thought into it as well and see if anything interesting comes to mind.

satanTime:
Hi guys,

I have 3 options:

**add a flag for the else case**
```typescript
class Test {
  @ValidateNested({ condition: isMySubClass })
  @IsString({ conditionElse: true }) // works in case if all `condition` returned false, then it fails.
  prop1: string | SubClass1;
}
```

**fail if all conditions failed and no other failures**
```typescript
class Test {
  @ValidateNested({ condition: isMySubClass })
  @IsString({ condition: isString })
   // always fails if all `condition` returned false.
  prop1: string | SubClass1;
}
```
can be too strict. For example in the next case:
```typescript
class Test {
  @IsNumber()
  @Min(300, { condition: o => o.something === true })
  prop1: string | SubClass1;
}
```

**a special decorator to fail if all conditions failed**
close to the first case, but more readable IMHO.
```typescript
class Test {
  @ValidateNested({ condition: isMySubClass })
  @IsString({ condition: isString })
  @RequireCondition() // works in case if all `condition` returned false, then it fails.
  prop1: string | SubClass1;
}
```

After different cases when we have conditions, when we don't have them, when we have mix, I would vote for the 3rd case with `@RequireCondition`, it looks better than `conditionElse` IMHO, and give us flexibility to say when we really want to fail in case when all conditions didn't match to trigger any validation.

davidthor:
Thanks for taking a look @satanTime! While I can see how consolidating around `condition` is ideal per your 3rd (and prefered) case, it does require a bit of verbosity to cover what may have otherwise been an _else_ or _default_ case. 

I wonder if there are any wholes you can find in a 4th proposal - conditional references superseding others and preventing them from running.

```typescript
class Test {
  @IsString({ condition: isString })
  @ValidateNested()
  prop1: string | SubClass1;
}
```

There's an obvious weakness in that it appears to prevent rules from being applied to the field regardless of conditions, but that would actually be covered as follows:

```typescript
class Test {
  @IsString({ condition: isString })
  @ValidateNested()
  @IsNotEmpty({ condition: () => true })  // preferably this would just be `condition: true`, but might require shorthand
  prop1: string | SubClass1;
}
```

So long as the conditions can match multiple, the lack of a condition can be interpreted as an `else` case. I'd love your thoughts!

satanTime:
Hi @davidthor,

I hope you're doing good.

To be honest, I didn't get fully `superseding others`, does it suspend all others, or only decorators without conditions and that returned false? 

I would say for me it feels like that
```typescript
  @IsString({ condition: isString }) // is executed only when isString returns true
  @ValidateNested() // is executed always, whatever isString returns
  @IsNotEmpty({ condition: () => true }) // the same as above - is executed always
  @SomeThingElse({ condition: true }) // the same as both above, doesn't feel like an else statement.
  field: any;
```

and if 
```typescript
  @IsNotEmpty()
  @IsString()
  @MinLength(5)
  field: any;
```
executes all validations, I would also expect
```typescript
  @IsNotEmpty()
  @IsString({condition: isString})
  @MaxLength({condition: isString})
  field: any;
```
to trigger all validators in case if `isString` returns `true`.
Because with superseding `''` becomes a valid case, when it feels like `IsNotEmpty` applies always and others in case of triggered conditions.

satanTime:
Hi guys, any update here this weekend?

satanTime:
Hi @vlapo and @davidthor, any update?

davidthor:
Hi @satanTime. I unfortunately haven't had time to re-address this, and it doesn't look like I have time in the coming week or two either. I'll close this for now and can potentially circle back at a later date.

satanTime:
Sad sad, I have time to work on it, only need mass approval on the final interface.

satanTime:
Hey @vlapo, I hope this finds you well, might you approve the approach?

joelcoronah:
Hi guys, @davidthor @satanTime, any update?

satanTime:
Hi @joelcoronah,

I'm waiting an approval from @vlapo and will start next moment.

NoNameProvided:
Hi, @satanTime can you open a new issue for this proposing the feature with some example showcasing how this would work?

satanTime:
Hi @NoNameProvided, 

this is in this thread: https://github.com/typestack/class-validator/pull/584#issuecomment-620681934

Please let me know if it fits your vision too and then I'll organise a new ticket for it for all the info.

NoNameProvided:
Reopening so this discussion wont go under the radar.

satanTime:
Hi @NoNameProvided,

what do you think about the proposal? If it works - I can start to work on the solution.

NoNameProvided:
I will need some time to think this over. I will get to it probably on the weekend.

satanTime:
Thanks, looking forward to your feedback.

ForkInSpace:
Waiting for this to be resolved as well. Much needed feature

satanTime:
Hi @NoNameProvided,

happy New Year! Any updates on the proposal? https://github.com/typestack/class-validator/pull/584#issuecomment-620681934

NoNameProvided:
Hi @davidthor. Sorry for the late reply. This is hopelessly conflicted. I cannot resolve it. If you are still interested in this PR (or anyone else) please rebase this or open a new one.

braaar:
See [my recent comment](https://github.com/typestack/class-validator/issues/1721#issuecomment-1857290706) in #1721

PR #587, opened 2020-04-22
chore: correct name in MaxLength validation decorator

jengel3:
Simple change, but this threw me off for a second.


vlapo:
Thank you

PR #590, opened 2020-04-24
build: replace mocha, sinon and chai with jest

rmuchall:
Hi @vlapo ,

As discussed here is the first of two PRs.
This one contains Jest only. If/when it is accepted I'll do the eslint PR :)

Please let me know if you have any comments/concerns.
Thanks,
Rob


.eslintrc.js
```diff
@@ -0,0 +1,33 @@
...
+        "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/no-unused-vars": "off",
+        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/member-delimiter-style": "off",
+        "@typescript-eslint/member-ordering": "error",
+        "@typescript-eslint/unbound-method": ["error", {
+            "ignoreStatic": true
+        }]
+    }
+};
```
vlapo, line 33: Lets do this in another PR :) Also applies to `tslint.json`

gulpfile.ts
```diff
@@ -72,14 +68,10 @@ export class Gulpfile {
      */
     @Task()
     compileTests() {
```
vlapo, line 70: `compileTests` should be `runTests`

gulpfile.ts
```diff
@@ -241,55 +233,12 @@ export class Gulpfile {
...
-            .pipe(gulp.dest("./coverage"));
-    }
-
     /**
      * Compiles the code and runs tests.
      */
     @SequenceTask()
     tests() {
-        return ["clean", "compileTests", "tslint", "coveragePre", "coveragePost", "coverageRemap"];
+        return ["clean", "compileTests"];
```
vlapo, line 241: Please do not remove `tslint` task from this sequence. We should force contributors to write tslint/eslint compliance code. Also I think it should be run before tests.

test/functional/custom-decorators.spec.ts
```diff
@@ -1,261 +1,241 @@
...
-        }
-
-        class MyClass {
-            firstName: string;
-
-            @IsShortenThan("firstName", {
-                message: "$property must be shorter then $constraint1. Given value: $value"
-            })
-            lastName: string;
+describe("decorator with separate validation constraint class", function () {
```
vlapo, line 175: Please use everywhere arrow functions `() => {` :)

test/functional/validation-options.spec.ts
```diff
@@ -20,1043 +12,1034 @@ const validator = new Validator();
 // Specifications: common decorators
 // -------------------------------------------------------------------------
 
-describe("validation options", function() {
```
vlapo, line 23: Why did you remove `describe`? I saw you remove root describe in some specs (but not at all of them). I would love to keep root describe.

test/functional/custom-decorators.spec.ts
```diff
@@ -1,261 +1,241 @@
...
-
-// -------------------------------------------------------------------------
-// Setup
-// -------------------------------------------------------------------------
-
 const validator = new Validator();
 
-// -------------------------------------------------------------------------
-// Specifications: common decorators
-// -------------------------------------------------------------------------
```
vlapo, line 24: Please remove or keep these comments in all spec files :)

.eslintrc.js
rmuchall: My bad, I included this file by mistake.

gulpfile.ts
rmuchall: This is not a change that I made.

gulpfile.ts
rmuchall: I removed tslint because it appeared to not be linting any files.

test/functional/custom-decorators.spec.ts
rmuchall: Yes, my mistake. I thought I had changed them all but missed this one.

test/functional/validation-options.spec.ts
rmuchall: Jest does not require root describe blocks unless you wish to include before/after setup for all tests within the block. As it was superfluous I removed it.

test/functional/custom-decorators.spec.ts
rmuchall: I will leave this one for you to fix ;)

rmuchall:
> I thought about promises. Maybe we should unite code style (at least in tests) and use callback style or async/await

I attempted to adopt a consistent style wherever possible through all the tests. Most tests I refactored to return a promise. Those that don't were already configured to use a helper function and it was quicker just to modify those to use expect();

rmuchall:
Thanks @vlapo :)
Thank you for your time and your patience.
I will do the eslint PR now.

RE: prettier
Prettier is not something that I use in my projects but I can try to install/configure it for class-validator if you wish it.

vlapo:
> RE: prettier
> Prettier is not something that I use in my projects but I can try to install/configure it for class-validator if you wish it.

Lets start with eslint. IMHO I do not like prettier format :-D maybe eslint is enough for us

rmuchall:
Yes, I feel the same as you. I do not like the prettier format either :)

PR #591, opened 2020-04-24
build: replace tslint with eslint

rmuchall:
Hi @vlapo,
Here is the PR for eslint as discussed :)
Thanks,
Rob


test/functional/validation-functions-and-decorators.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
vlapo, line 1: Please do not disable rules globally. Use `eslint-disable-next-line`.

src/validation/Validator.ts
```diff
@@ -10,65 +10,41 @@ import * as validator from "validator";
...
-
     // -------------------------------------------------------------------------
     // Public Methods
     // -------------------------------------------------------------------------
 
     /**
      * Performs validation of the given object based on decorators used in given object class.
      */
-    validate(object: Object, options?: ValidatorOptions): Promise<ValidationError[]>;
+    validate(object: Record<string, any>, options?: ValidatorOptions): Promise<ValidationError[]>;
```
vlapo, line 20: Are we sure `Record<string, any>` is same as `Object`? This is change in our API so I want to be sure.
I think it is ok but I want extra check. @satanTime could you please check this?

.eslintrc.js
```diff
@@ -0,0 +1,33 @@
...
+        "no-prototype-builtins": "off",
+        "@typescript-eslint/no-inferrable-types": "off",
+        "@typescript-eslint/no-non-null-assertion": "off",
+        "@typescript-eslint/no-unused-vars": "off",
+        "@typescript-eslint/no-explicit-any": "off",
+        "@typescript-eslint/member-delimiter-style": "off",
+        "@typescript-eslint/member-ordering": "error",
+        "@typescript-eslint/unbound-method": ["error", {
+            "ignoreStatic": true
+        }]
```
vlapo, line 31: I think we should remove `"@typescript-eslint/member-delimiter-style": "off",` and `"no-case-declarations": "off",`

src/validation/Validator.ts
satanTime: almost, but key can also be a number and a symbol, the best way would be `keyof any` instead of `stream`. I'm checking the changes and prepare a feedback soon.

satanTime: The error message is:
```
class-validator/src/validation/Validator.ts
  37:30  error  Don't use 'Object' as a type. Use Record<string, any> instead  @typescript-eslint/ban-type
```

but the definition of `Record` is
```typescript
/**
 * Construct a type with a set of properties K of type T
 */
type Record<K extends keyof any, T> = {
    [P in K]: T;
};
```

To be honest I don't know people who are using objects with number / symbol keys, but if it's supported in an object would be good to see it in the lib too.
```typescript
Record<keyof any, any>
```

src/decorator/array/ArrayContains.ts
```diff
@@ -7,11 +7,11 @@ export const ARRAY_CONTAINS = "arrayContains";
  * Checks if array contains all values from the given array of values.
  * If null or undefined is given then this function returns false.
  */
-export function arrayContains(array: unknown, values: any[]) {
+export function arrayContains(array: unknown, values: any[]): boolean {
     if (!(array instanceof Array))
         return false;
 
-    return values.every(value => array.indexOf(value) !== -1);
+    return values.every(value => array.includes(value));
```
satanTime, line 14: @vlapo, is class-validator used on frontend? `includes` requires a polyfill in IE.

satanTime: the rest of the PR looks good.

test/functional/validation-functions-and-decorators.spec.ts
satanTime: Because it's a test, I would say it even can be removed and the names of keys can be fixed.

src/validation/Validator.ts
vlapo: Maybe use `object` (lowercase first "o") - https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type

test/functional/validation-functions-and-decorators.spec.ts
vlapo: We should use camelcase everywhere in our code. This break rule is only because of validatorjs interfaces so we should disable only lines where validatorjs snake case style is used.

src/validation/Validator.ts
satanTime: IMO works too, a replacement of `Object` with `{[key: keyof any]: any}` or `Record` etc. for me is just a hotfix of linting and in case of general libraries such as class-validator to be precise isn't so easy.
As I understand from documentation `Object` is kind of `any` because it matches primitives when `object` is an object as we understand it https://take.ms/RYUk6
if we expect a normal object - then `object` is the perfect match considering the `null` case.

src/decorator/array/ArrayContains.ts
vlapo: Yeah we should support browsers too. Unfortunately we do not have tests for this and also we do not have define minimal supported browser versions. 

Maybe we can use eslint plugin - https://github.com/amilajack/eslint-plugin-compat

src/validation/Validator.ts
vlapo:
```ts
declare function original(o: Object): void;
declare function objectType(o: object): void;
declare function record(o: Record<any, any>): void;

class TestClass { }

// ok case
original({ a: 1, 1: 2 }); // OK
objectType({ a: 1, 1: 2 }); // OK
record({ a: 1, 1: 2 }); // OK

original(new TestClass()); // OK
objectType(new TestClass()); // OK
record(new TestClass()); // OK

original(Object.create({})); // OK
objectType(Object.create({})); // OK
record(Object.create({})); // OK

original(Object.create(null)); // OK
objectType(Object.create(null)); // OK
record(Object.create(null)); // OK

original(null); // Error
objectType(null); // Error
record(null); // Error

original(undefined); // Error
objectType(undefined); // Error
record(undefined); // Error

original(Symbol("key")); // OK
objectType(Symbol("key")); // Error
record(Symbol("key")); // Error

original(true); // OK
objectType(true); // Error
record(true); // Error

original('asd'); // OK
objectType('asd'); // Error
record('asd'); // Error

original(1); // OK
objectType(1); // Error
record(1); // Error

original(() => false); // OK
objectType(() => false); // OK
record(() => false); // OK
```
Looks like `Record<any, any>` is equivalent for `object`. However this looks like public API change (we should move it into 0.13.0 release) **from my point of view is this fix because we want only objects as input for validation 😄**

satanTime: agree, make sense. For me Record is too long to say any object. Anyway `Object` doesn't support `null`, so `object` looks like the best match.

test/functional/validation-functions-and-decorators.spec.ts
rmuchall: Resolved.

src/validation/Validator.ts
rmuchall: Changed Record<string, any> to object.

.eslintrc.js
rmuchall: Resolved.

src/decorator/array/ArrayContains.ts
rmuchall: https://caniuse.com/#search=includes

rmuchall: This is governed by the following eslint rule:
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-includes.md

vlapo: Let's add https://github.com/amilajack/eslint-plugin-compat plugin and define config like `"browserslist": ["> 5%"],`

Also disable `prefer-includes` please.

src/validation/Validator.ts
rmuchall: I'd like to reopen the discussion on Record<string, any> vs object :)

There are some subtle differences that may impact class-validator.
object is a primitive type, which means we cannot use it with generics

Perhaps Record<string, any> is the correct usage here?

```
declare function genericRecord<T extends Record<string, any>>(o: T): void;
declare function genericObject<T extends object>(o: T): void;

function genericRecordWrapper<T>(o: T): void {
    genericRecord(o); // OK
}

function genericObjectWrapper<T>(o: T): void {
    genericObject(o); // ERROR
   // Argument of type 'T' is not assignable to parameter of type 'object' (2345)
}
```

rmuchall: perhaps Record<string | Symbol, any> is better?

satanTime: the right type for keys is `keyof any`, but your point is interesting, I'll check it deeper tonight too.

rmuchall: Are you sure?
From the Moz link above:
"Property names are string or Symbol. Any other value, including a number, is coerced to a string."

satanTime: Moz is about javascript when we discuss typescript here and in its case number and string are different types. Check source code of Record and you can see `keyof any` there. I would rely on knowledge of typescript's team.

rmuchall: Ah, good point. I understand. So perhaps Record<any, any> is the best solution?

rmuchall: I've been thinking about this some more and I've realized that I was wrong.
object = any non primitive type
So my earlier code was wrong/badly typed.
```
declare function genericRecord<T extends Record<string, any>>(o: T): void;
declare function genericObject<T extends object>(o: T): void;

function genericRecordWrapper<T>(o: T): void {
    genericRecord(o); // OK
}

// This is now a correct definition. It does not allow primitive types
function genericObjectWrapper<T extends object>(o: T): void {
// function genericObjectWrapper<T>(o: T): void {
    genericObject(o); // OK
}
```
So object *is* correct usage for class-validator :)

satanTime: Ah, call, I almost started the investigation. Then glad to hear and looking forward to see your changes in merged :)

src/decorator/array/ArrayContains.ts
vlapo: @rmuchall could you please work in this changes?

rmuchall: Yes, apologies for the delay. I'll finish them off tomorrow :)

vlapo: Its okey. Just friendly reminder :)

rmuchall: Done :)

src/.eslintrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+    env: {
+        browser: true,
+    },
+    extends: [
+    	"plugin:compat/recommended"
```
vlapo, line 6: Why did you create another eslintrc?

package.json
```diff
@@ -68,14 +73,14 @@
...
     "tslib": "^1.11.1",
-    "tslint": "^5.11.0",
-    "tslint-stylish": "^2.1.0",
     "typescript": "^3.5.3",
     "webpack-config-utils": "^2.3.1"
   },
   "scripts": {
     "build": "gulp package",
     "test": "gulp tests",
+    "eslint": "gulp eslint",
```
vlapo, line 82: This name should be `lint`. In case we change lint tool in future :)

src/.eslintrc.js
rmuchall: So that /src is linted separately from /tests.
Tests don't require browser compatibility.

package.json
rmuchall: OK :)

rmuchall: Done.

src/.eslintrc.js
vlapo: Oh, I see now. Thnx

PR #600, opened 2020-04-30
feat: allow readonly array argument for IsIn IsNotIn validators

### LINKED ISSUE #596: Ability to use constant array as `IsIn` and `IsNotIn` argument
Constant array can not be passed as `IsIn` argument, which is a common thing to validate against a constant array not a dynamic list. Example:

```
const Currencies = ['ETH', 'BTC'] as const;

class Payment {
  @IsIn(Currencies)
  currency: string
}
```

Generates compilation error:

```
Argument of type 'readonly ["ETH", "BTC"]' is not assignable to parameter of type 'any[]'. The type 'readonly ["ETH", "BTC"]' is 'readonly' and cannot be assigned to the mutable type 'any[]'
```

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions
Fixable by using `readonly any[]` as argument types instead of `any[]`.

bogdan:
Fixed #596.

### DIFF
--- a/src/decorator/common/IsIn.ts
+++ b/src/decorator/common/IsIn.ts
@@ -6,14 +6,14 @@ export const IS_IN = "isIn";
 /**
  * Checks if given value is in a array of allowed values.
  */
-export function isIn(value: unknown, possibleValues: unknown[]): boolean {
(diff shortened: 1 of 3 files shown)


vlapo:
Thank you

Issue #399, opened 2019-08-13
Validation order


lukejagodzinski:
Can I define validation order somehow? I haven't found anything in docs and issues

gremo:
It seems reversed. And I think it should stop at the first error:

```
"constraints": {
"length": "abiCode must be longer than or equal to 5 characters",
"isNotEmpty": "abiCode should not be empty"
}
```

For me, it doesn't make any senso to return "length" constraint violation. At least, it should be configurable (`stopAtFirstError`)

lukejagodzinski:
@gremo where do you use the `constraints` option? In the validator decorator?

gremo:
@lukejagodzinski nope, that is the output of `validate `method. It show that the validation order is reversed.

peresbruno:
Same question here.

forsigner:
Same question here +1

Demonell:
That would be nice option, indeed

valdineifer:
> @gremo I didn't ask about in what order validation was performed but about how to define custom order

Wow, this sounds rude.
The @gremo was right. I change the order of the validation decorators and it worked.

Issue #188, opened 2018-04-13
feat: add option to stop validation after first validation error


HonoluluHenk:
If i apply the following annotations:

```ts
  @IsNotEmpty()
  @IsInt()
  @Min(0)
  foobar: number;
```

The validation result will be:

```json
{ 
  "value": null, 
  "property": "foobar", 
  "children": [], 
  "constraints": {
    "min": "foobar must be greater than 0",
   "isInt": "foobar must be an integer number", 
   "isNotEmpty": "foobar should not be empty" 
  } 
}
```

If no value is given (i.e.: `foobar === null`), I only want the validation result of `@IsNotEmpty`.
All the others are quite useless... worse: since the validation errors (the constraints) are stored in an object, the ordering is lost so I cannot "just take the first error constraint" :(

Expected result:
```
{ 
  "value": null, 
  "property": "foobar", 
  "children": [], 
  "constraints": {
   "isNotEmpty": "foobar should not be empty" 
  } 
}
```

or:

```
 { 
  "value": null, 
  "property": "foobar", 
  "children": [], 
  "constraints": [
    {"isNotEmpty": "foobar should not be empty" },
    {"isInt": "foobar must be an integer number"},
    { "min": "foobar must be greater than 0"}
  ]
}
```

NoNameProvided:
Is this a feature request? What would be your ideal format?

HonoluluHenk:
Yes, this is a feature request :)

The result with all validations as list would be nice:
```
{ "value": null,
  "property": "foobar",
  "children": [], 
  "constraints": [ 
    {"isNotEmpty": "foobar should not be empty" },
    {"isInt": "foobar must be an integer number"},
    { "min": "foobar must be greater than 0"}
}
```
with the constraints ordering same as annotation ordering

kostanev:
It would be nice if there is a setting that makes the validator stop after the very first fail. For example, I have custom constraint that checks the database for email existence. I used the following decorators:
```
@IsNotEmpty()
@Length(3, 10)
@EmailNotExists()
email: string;
```
This makes database query even if email is empty or with less than 3 characters. Of course I could add additional checks in my constraint, but it makes no sense to do that.

HonoluluHenk:
@kostanev is making a good point

HonoluluHenk:
Regarding: making the validator stop after the very first fail...

I just had a look a Java Bean Validation: https://docs.oracle.com/javaee/6/tutorial/doc/gircz.html
[...]

Issue #148, opened 2017-12-14
Feature request: Fail on first ValidationError.


janis91:
It would be very nice to have a configuration option like 'failFast' or something like that available. It makes testing much easier and in case of big validation tasks it also gives a big performance boost:
if something is wrong it will fail fast and doesn't have to wait for the rest of the validation.

What do you think?

NoNameProvided:
Sound reasonable to me. :)

NoNameProvided:
Closing in favor of #178

PR #620, opened 2020-05-27
feat: add option to stop validation at first error

ytetsuro:
## Description

Added an option to not detect more than one Error for each item.

## Checklist

  - `npm run lint:check` passes (An error occurs outside of this PR.)

### Fixes

references #148, references #399, references #188

### DIFF
--- a/src/validation/ValidatorOptions.ts
+++ b/src/validation/ValidatorOptions.ts
@@ -59,4 +59,9 @@ export interface ValidatorOptions {
    * Settings true will cause fail validation of unknown objects.
    */
   forbidUnknownValues?: boolean;
+
+  /**
+   * Settings true will stop validation when encount first error.
+   */
+  stopAtFirstError?: boolean;
 }
--- a/src/validation/ValidatorOptions.ts
+++ b/src/validation/ValidatorOptions.ts
@@ -61,7 +61,7 @@ export interface ValidatorOptions {
   forbidUnknownValues?: boolean;
 
   /**
-   * Settings true will stop validation when encount first error.
+   * When set to true, validation of the given property will stop after encountering the first error. Defaults to false.
    */
   stopAtFirstError?: boolean;
 }
--- a/test/functional/validation-options.spec.ts
+++ b/test/functional/validation-options.spec.ts
@@ -1090,10 +1090,10 @@ describe('context', () => {
   it('should stop at first error.', () => {
     class MyClass {
       @IsDefined({
-        message: 'isDefined'
+        message: 'isDefined',
       })
       @Contains('hello', {
-        message: 'String is not valid. You string must contain a hello word'
+        message: 'String is not valid. You string must contain a hello word',
       })
       sameProperty: string;
     }
--- a/test/functional/validation-options.spec.ts
+++ b/test/functional/validation-options.spec.ts
@@ -1087,29 +1087,29 @@ describe('context', () => {
     });
   });
 
-  it("should stop at first error.", () => {
+  it('should stop at first error.', () => {
     class MyClass {
       @IsDefined({
-        message: "isDefined",
+        message: 'isDefined',
         context: {
-          bye: "now"
-        }
+          bye: 'now',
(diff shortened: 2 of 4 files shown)


molinab297-unisys:
If we could get this merged that would be awesome... #188 shows how much demand there is for this feature.

ytetsuro:
@NoNameProvided

A conflict was occurring, so I rebased it.

However, CI/CD is giving me an error at a point completely unrelated to this fix.
Does this need to be fixed in this PR?

src/validation/ValidatorOptions.ts
```diff
@@ -59,4 +59,9 @@ export interface ValidatorOptions {
    * Settings true will cause fail validation of unknown objects.
    */
   forbidUnknownValues?: boolean;
+
+  /**
+   * Settings true will stop validation when encount first error.
```
NoNameProvided, line 64:
```suggestion
   * When set to true, validation will return after encountering the first error. Defaults to false.
```

test/functional/validation-options.spec.ts
```diff
@@ -1086,4 +1086,30 @@ describe('context', () => {
...
+    class MyClass {
+      @IsDefined({
+        message: 'isDefined',
+        context: {
+          bye: 'now',
+        },
+      })
+      @Contains('hello', {
+        message: 'String is not valid. You string must contain a hello word',
+        context: {
```
NoNameProvided, line 1100: Checking context functionality is not part of this test, please remove and add to where context is tested.

test/functional/validation-options.spec.ts
```diff
@@ -1086,4 +1086,30 @@ describe('context', () => {
...
+    }
+
+    const model = new MyClass();
+    return validator.validate(model, { stopAtFirstError: true }).then(errors => {
+      expect(errors.length).toEqual(1);
+      expect(Object.keys(errors[0].constraints).length).toBe(1);
+      expect(errors[0].contexts['isDefined']).toEqual({ bye: 'now' });
+      expect(errors[0].constraints['isDefined']).toBe('isDefined');
+    });
+  });
```
NoNameProvided, line 1114: Also, add a test here when `stopAtFirstError` is set to true. (I know this is the default behavior and have a lot of implicit test for this, but let's test it explicitly too.

ytetsuro: I am writing here to test when `stopAtFirst` is set to true.
Are you saying that we should also test for false in this test, as shown below?

```ts
return validator.validate(model, { stopAtFirstError: true }).then(errors => {
      expect(errors.length).toEqual(1);
      expect(Object.keys(errors[0].constraints).length).toBe(1);
      expect(errors[0].contexts['isDefined']).toEqual({ bye: 'now' });
      expect(errors[0].constraints['isDefined']).toBe('isDefined');
    }).then(() => validator.validate(model, { stopAtFirstError: false })).then(errors => {
      expect(Object.keys(errors[0].constraints).length).toBe(2);
    });
```

src/validation/ValidatorOptions.ts
ytetsuro: fixed.

test/functional/validation-options.spec.ts
ytetsuro: fixed.

test/functional/validation-options.spec.ts
ytetsuro: @NoNameProvided

NoNameProvided: Yes, usually it's good to explicitly test the opposite to show the intent. (We have a lot of implicit test for the default behavior, but it's always better to be explicit. For every feature behind flag there should be a test for on and off version.)

I will go ahead and merge it as it is, in future contribution please keep this in mind. Thanks!

ytetsuro: Yes, sir.

ytetsuro:
Thanks merged.

PR #634, opened 2020-06-11
fix: allow "any" as locale in isPostalCode decorator

main-kun:
PostalCodeLocale in @types/validator.js does not include 'any' but isPostalCode should be able to receive it as a locale

[PostalCodeLocale in @types/validator.js ](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/80922190a8125f052351520d07cb5166fc9b917b/types/validator/index.d.ts#L957)


NoNameProvided:
Please instead of merge, rebase your branch to the develop, otherwise looks good!

NoNameProvided:
Thanks!

PR #635, opened 2020-06-12
feat: add nullable option for IsNotEmptyObject rule.

ytetsuro:
_Note: PR description was moved to issue for easier discoverability by @NoNameProvided ._

references #678.

### DIFF
--- a/src/decorator/object/IsNotEmptyObject.ts
+++ b/src/decorator/object/IsNotEmptyObject.ts
@@ -8,15 +8,21 @@ export const IS_NOT_EMPTY_OBJECT = 'isNotEmptyObject';
  * Checks if the value is valid Object & not empty.
  * Returns false if the value is not an object or an empty valid object.
  */
-export function isNotEmptyObject(value: unknown): boolean {
-  if (!isObject(value)) {
-    return false;
-  }
-  for (const key in value) {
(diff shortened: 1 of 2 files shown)


NoNameProvided:
Hi @ytetsuro! 

This looks good, however the decorator signature should be `IsNotEmptyObject({ nullable: boolean })`. 

Can you rebase this to the current develop or create a new PR for this? (You will need to update the tests as we moved from Mocha to Jest.

ytetsuro:
> This looks good, however the decorator signature should be IsNotEmptyObject({ nullable: boolean }).

> Can you rebase this to the current develop or create a new PR for this? (You will need to update the tests as we moved from Mocha to Jest.

OK.
Thanks, good solution response.

ytetsuro:
@NoNameProvided 
fixed.

Please re-review.

NoNameProvided:
Thanks! 

PS: In future please don't use `it.each` for tests as it tends to make things harder to read.

ytetsuro:
Thanks merged.

@NoNameProvided

> PS: In future please don't use it.each for tests as it tends to make things harder to read.

Yes, sir.
But why?

NoNameProvided:
In unit tests, I like to keep things as simple as possible, and the `it.each` format makes things harder to read.

PR #642, opened 2020-06-19
fix: matches decorator with modifier as string

### LINKED ISSUE #626: fix: @Matches() decorator receives parameters in wrong order
Oh hi there! 😄

I think it is a bug, but I might be doing something wrong.

### Description

See the following class:

```javascript
class Example {
    @Matches('^[a-z]+[A-Z][a-z]*$', 'g', {
       each: true,
       message: v => `${v.value} is illegal in ${v.property}`
    })
    @IsArray()
    types: string[];
}
```

I got error stack like this

```
error: SyntaxError: Invalid flags supplied to RegExp constructor '^[a-z]+[A-Z][a-z]*$'
        at new RegExp (<anonymous>)
```

Then read the Matches decorators code, and find a [matches function](https://github.com/typestack/class-validator/blob/d74518169e12f5e47d839bda9066b90a34947de6/src/decorator/string/Matches.ts#L13) is called by [Matches decorators function](https://github.com/typestack/class-validator/blob/d74518169e12f5e47d839bda9066b90a34947de6/src/decorator/string/Matches.ts#L36) 

```javascript
export function Matches(pattern: RegExp | string, modifiersOrAnnotationOptions?: string | ValidationOptions, validationOptions?: ValidationOptions): PropertyDecorator {
    ......
    return ValidateBy(
        {
            ......
            validator: {
                validate: (value, args) => matches(value, args.constraints[0], args.constraints[0]),
                ......
            }
        },
        validationOptions
    );
}
```

```javascript
export function matches(value: string, pattern: RegExp | string, modifiers?: string): boolean {
    return typeof value === "string" && validator.matches(value, pattern as unknown as any, modifiers);
}
```
I'm confused why use `matches(value, args.constraints[0], args.constraints[0])`, this like pattern equal to modifiers, when `validator.matches` is executed, `new RegExp(pattern, modifiers)` throw that SyntaxError.

I might be thinking something wrong

If someone could have a look it would be appreciated.

Cheers

### Environment

- [x] nodejs: 12
- [ ] browser: 
- [ ] framework/library: 

**class-validator version:** <code>0.12.2</code>

Kiliandeca:
fixes #626

### DIFF
--- a/src/decorator/string/Matches.ts
+++ b/src/decorator/string/Matches.ts
@@ -37,7 +37,7 @@ export function Matches(
       name: MATCHES,
       constraints: [pattern, modifiers],
       validator: {
-        validate: (value, args): boolean => matches(value, args.constraints[0], args.constraints[0]),
+        validate: (value, args): boolean => matches(value, args.constraints[0], args.constraints[1]),
         defaultMessage: buildMessage(
           (eachPrefix, args) => eachPrefix + '$property must match $constraint1 regular expression',
           validationOptions
--- a/test/functional/validation-functions-and-decorators.spec.ts
+++ b/test/functional/validation-functions-and-decorators.spec.ts
@@ -3706,7 +3706,7 @@ describe('MaxLength', () => {
   });
 });
 
-describe('Matches', () => {
+describe('Matches pattern RegExp', () => {
   const constraint = /abc/;
   const validValues = ['abc', 'abcdef', '123abc'];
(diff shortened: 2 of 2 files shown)


NoNameProvided:
Hi! 

Can you rebase these changes to the current develop?

Kiliandeca:
@NoNameProvided it's done

NoNameProvided:
Thanks!

PR #662, opened 2020-07-25
Fix/accept unknown types

yktakaha4:
I'm helped by this product a lot.

Some validation methods accepted only limited types, but I think it should accept unknown types like `isEmail`.
(Confirmed to pass existing test.)

https://github.com/typestack/class-validator/blob/1955250de64d76504df465aa02076a2e5b219bb7/src/decorator/string/IsEmail.ts#L11-L13


NoNameProvided:
Closing this as wontfix. 

If you still experience issues related to the current behavior please open a question and detail your use-case. Thanks.

PR #681, opened 2020-08-08
test: return async assertions

blaugold:
Fixes some test which should return a promise, but instead left the promises unhandled.


blaugold:
Closed in favor of #683.

PR #682, opened 2020-08-08
build: setup project for TypeScript solution file

blaugold:
Configures prod, spec and sample projects and references them in solution file.
This improves integration with tooling such as IDEs.
See [TS docs](https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure) for guidance on how to use project references to structure TS projects.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -4,9 +4,10 @@ plugins:
 parserOptions:
   ecmaVersion: 2018
   sourceType: module
-  project: 
-    - ./tsconfig.json
+  project:
+    - ./tsconfig.prod.json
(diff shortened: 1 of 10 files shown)


NoNameProvided:
Thanks for this contribution, but I will close this as wontfix. 

This as just an extra layer of complexity. We already have inheritance support, between the configurations and the current solution is as lean as we can get. 

The default tsconfig is stricter on purpose than the build, so IDEs will report errors but compilation will work and we can release.

To avoid extra work and rejected changes, please open an issue before sending a patch witch is not a bugfix so we can discuss it before implementation. 

PS: Please use the PR template and follow the instructions.

PR #683, opened 2020-08-08
test: fix errors and enable linting of tests

blaugold:
This PR adds the `test` dir to the lint scripts in `package.json` and fixes current lint errors.

Incidentally there were a few async test which did not return a promise, possibly leading to false positives.
Two of these tests should have been failing and are fixed as part of this PR.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -4,7 +4,7 @@ plugins:
 parserOptions:
   ecmaVersion: 2018
   sourceType: module
-  project: 
+  project:
     - ./tsconfig.json
     - ./tsconfig.spec.json
 extends:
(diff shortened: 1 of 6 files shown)


NoNameProvided:
Hi! 

Thanks for the fix regarding the returns! About the other changes: I am kind of conflicted about enabled lint in the test folder as we will just end up disabling rules everywhere when the test requires (like you did in multiple places). 

Anyway, let's try this but then please disable the `@typescript-eslint/restrict-template-expressions` rule altogether because that is a useless rule.

blaugold:
I agree, tests don't need to be quite as clean as production code. I just noticed that the linter caught some bugs like duplicate test names and the mentioned async asserts.

`@typescript-eslint/restrict-template-expressions` is now `off`.

PR #799, opened 2020-10-31
fix: correct typo in IsUUID decorator

BJvdA:
## Description
Fix typo in IsUUID.js error message, it should be `a UUID` instead of `an UUID`, since UUID is pronounced with a hard Y. Just like UFO.


NoNameProvided (approved):

NoNameProvided:
Thanks!

PR #830, opened 2020-12-07
feat: add optional matcher function for ArrayUnique decorator

### LINKED ISSUE #602: ArrayClassUnique, or property-based uniqueness testing
I would be picturing something like

``` typescript
@ArrayClassUnique(User, (u:User) => u.name)
users:User[]
```

with

``` typescript
function arrayClassUnique2(array: unknown, args: ValidationArguments): boolean {
    if (!(array instanceof Array))
        return false;

    const constructor = args.constraints[0];
    const fn = args.constraints[1];
    const ids = array.map((v: unknown) => {
        if (isInstance(v, constructor)) {
            return fn(v);
        } else {
            return null;
        }
    });

    const uniqueItems = ids.filter((a, b, c) => { return c.indexOf(a) === b; });

    return array.length === uniqueItems.length;
}

export function arrayClassUnique(value: unknown, args: ValidationArguments): boolean {

    if (!(value instanceof Array))
        return false;

    const constructor = args.constraints[0];
    const fn = args.constraints[1];
    const ids = value.map((v: unknown) => {
        if (isInstance(v, constructor)) {
            return fn(v);
        } else {
            return null;
        }
    });
    const uids = new Set(ids);

    return uids.size == value.length
        && uids.has(null) == false
        && uids.has(undefined) == false;
}

/**
 * Checks if all array's values are unique. Comparison for objects is reference-based.
 * If null or undefined is given then this function returns false.
 */
export function ArrayClassUnique<T>(
    targetType: new (...args: unknown[]) => T,
    uidFn: (T) => Primitive,
    validationOptions?: ValidationOptions): PropertyDecorator {
    return ValidateBy(
        {
            name: ARRAY_CLASS_UNIQUE,
            constraints: [targetType, uidFn],
            validator: {
                validate: (value, args): boolean => { return arrayClassUnique(value, args); },
                defaultMessage: buildMessage(
                    (eachPrefix) => { return eachPrefix + 'All $property\'s elements must be unique'; },
                    validationOptions
                )
            }
        },
        validationOptions
    );
}
```

Type checking is a bit overboard, and should probably be done using IsInstance, but it allowed for stronger typing inside the lambda function. If it is kept, it should probably have its own message.

Diluka:
when identifer function supplied,
the comparison will base on the mapped array.

close #602

## Checklist

  - `npm run lint:check` passes

### Fixes


src/decorator/array/ArrayUnique.ts
```diff
@@ -2,14 +2,19 @@ import { ValidationOptions } from '../ValidationOptions';
...
 /**
  * Checks if all array's values are unique. Comparison for objects is reference-based.
  * If null or undefined is given then this function returns false.
  */
-export function arrayUnique(array: unknown): boolean {
+export function arrayUnique(array: unknown[], identifier?: ArrayUniqueIdentifier): boolean {
   if (!(array instanceof Array)) return false;
 
+  if (identifier) {
+    array = array.map(o => (o != null ? identifier(o) : o));
```
NoNameProvided, line 15: Why do you check for `null` here before calling the `identifier` function?

Diluka: check `null` before calling or check `null` in every `identifier` function. I just pick the former

PR #846, opened 2021-01-04
fix: do not throw in ValidationError.toString() error when using forbidNonWhitelisted option

### LINKED ISSUE #604: ValidationError.toString() breaks when forbidNonWhitelisted is enabled in ValidatorOptions
### Description

The [toString()](https://github.com/typestack/class-validator/blob/master/src/validation/ValidationError.ts#L51) method in the validation error class invokes the `map()` method on the `children` property (lines 60 and 70 at commit 1955250). However, this property cannot be guaranteed to be set in all cases. When whitelist and forbidNonWhitelisted are enabled concurrently, the `children` array is set to `undefined` [here](https://github.com/typestack/class-validator/blob/master/src/validation/ValidationExecutor.ts#115). If the `toString()` method is then invoked, a "Cannot ready property 'map' of undefined" error is thrown.

Proposed solution:
Since I don't think the `children` property serves a purpose for a whitelisting error, could the property be updated to be optional in the ValidationError class and a null check added to the `toString()` method before calling `this.children.map()`?

### Reproduction

```typescript
import 'reflect-metadata';

import { validateSync } from 'class-validator';

class Test {
  property: string;
}

const t = new Test();
t.property = "This property has no decorators and should fail validation."

/**
 * Errors are handled as expected since whitelisting is enabled and `property` does not
 * include the @Allow decorator. This produces a single error in the returned array.
 */
const errors = validateSync(t, { whitelist: true, forbidNonWhitelisted: true });
console.log(errors);

/**
 * However, once we try to call the `toString()` method, an error will be thrown since
 * the ValidationError in `errors` doesn't contain a `children` property.
 */
for (const e of errors) {
  console.log(errors.toString());
}
```

### Environment

- [ ] nodejs: 12.16.1
- [ ] browser: 
- [ ] framework/library: 

**class-validator version:** <code>0.12.2</code>

simon-a-j:
## Description

When validation is performed with the `forbidNonWhitelisted` option, for any non-whitelisted properties, a `ValidationError` is created with the `children` property `undefined`. This breaks the current implementation of `ValidationError.toString()` which will try to map the `children` property, causing an error to be thrown. See #604.

This PR resolves the issue by testing `children` before mapping. Keeping the property optional feels semantically preferable to an empty list. The whitelist validation test is updated to cover this scenario.

## Checklist

- [X] the pull request title describes what this PR does (not a vague title like `Update index.md`)
- [X] the pull request targets the *default* branch of the repository (`develop`)
- [X] the code follows the established code style of the repository
  - `npm run lint:check` passes
- [X] tests are added for the changes I made (if any source code was modified)
- [N/A] documentation added or updated
- [X] I have run the project locally and verified that there are no errors

### Fixes

fixes #604

### DIFF
--- a/src/validation/ValidationError.ts
+++ b/src/validation/ValidationError.ts
@@ -31,7 +31,7 @@ export class ValidationError {
   /**
    * Contains all nested validation errors of the property.
    */
-  children: ValidationError[];
+  children?: ValidationError[];
 
   /*
    * A transient set of data passed through to the validation result for response mapping
@@ -60,7 +60,9 @@ export class ValidationError {
           this.target ? this.target.constructor.name : 'an object'
         }${boldEnd} has failed the validation:\n` +
         (this.constraints ? propConstraintFailed(this.property) : ``) +
-        this.children.map(childError => childError.toString(shouldDecorate, true, this.property)).join(``)
+        (this.children
+          ? this.children.map(childError => childError.toString(shouldDecorate, true, this.property)).join(``)
+          : ``)
       );
     } else {
       // we format numbers as array indexes for better readability.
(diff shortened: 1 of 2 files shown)


NoNameProvided:
Thanks!

PR #940, opened 2021-03-01
feat: add support for phone number validation strategy

### LINKED ISSUE #939: feature: allow relying on `libphonenumber`'s `isPossible` for phone numbers validation
### Description

The phone numbers validation relies on `libphonenumber`'s [`isValid`](https://github.com/ruimarinho/google-libphonenumber/blob/master/src/phonenumberutil.js#L3059) function. This function operates a strict validation, but there's also the [`isPossible`](https://github.com/ruimarinho/google-libphonenumber/blob/master/src/phonenumberutil.js#L3356) function that basically only checks the number's format.

### Proposed solution

I'd like to be able to pass some options to the `IsPhoneNumber` decorator to make it rely on `isPossible`.

The best, non-breakin, approach would be for the decorator to accept either a `CountryCode` (current behavior) or a `{ region?: CountryCode; strategy: 'valid' | 'possible'; }` argument. The default behavior wouldn't change.

fwoelffel:
## Checklist

  - `npm run lint:check` passes

### Fixes

fixes #939


src/decorator/string/IsPhoneNumber.ts
```diff
@@ -4,18 +4,30 @@ import { parsePhoneNumberFromString, CountryCode } from 'libphonenumber-js';
 
 export const IS_PHONE_NUMBER = 'isPhoneNumber';
 
+type Options = { region?: CountryCode; strategy?: 'valid' | 'possible' };
+
+type RegionOrOptions = CountryCode | Options;
+
+function normalizeOptions(regionOrOptions?: RegionOrOptions): Options | undefined {
+  return 'string' === typeof regionOrOptions ? { region: regionOrOptions } : regionOrOptions;
```
braaar:
```suggestion
  return typeof regionOrOptions === 'string' ? { region: regionOrOptions } : regionOrOptions;
```
I would say this is the preferred syntax for this condition.

fwoelffel: Sure. I tend to use [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) but I understand these may not be to everybody's taste.

NoNameProvided:
Thanks for opening this PR, but this won't get merged. 

I don't think the pros outweigh the cons:
- I believe the vast majority of validation use cases check for valid phone numbers only
- it introduces a custom option that doesn't exist in the source lib and I prefer having all validators run with as little custom config as possible
- we actually changed to check for letters in phone numbers and fail the check and this is not possible with this approach

If you need this you can always create a custom decorator and use that.

PR #962, opened 2021-03-11
feat: Pass extra validation arguments to message builders

### LINKED ISSUE #961: feature: allow to pass extra arguments to message builders
### Description
It would be convenient to allow callers of ``validate()`` to pass extra arguments to message builders.

This feature can be used, for example, to pass i18n context of the caller in order to localize error messages:

```typescript
class MyClass {
  @IsNotEmpty({ message: ({ t }) => t('title.not.empty.key') })
  title: string = '';
}

const t = useTranslation(); // i18next

validate(..., { extraArguments: { t }}).then(...)
```

### Proposed solution
The proposed PR implements this feature by introducing ``ExtraValidationArguments`` type that can be augmented via type merging:

```typescript
 declare module 'class-validator' {
   interface ExtraValidationArguments {
     t: TFunction;
   }
 }
 ```

This approach doesn't tie ``class-validator`` to any particular 3rd party library while allowing users to pass extra arguments they need in type-safe way.

axtavt:
## Description
Allow caller of ``validate()`` to pass extra arguments to message builders.

## Checklist

  - `npm run lint:check` passes

### Fixes
fixes #961

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,7 @@ export interface ValidatorOptions {
 
   forbidUnknownValues?: boolean;
   stopAtFirstError?: boolean;
+  extraArguments?: ExtraValidationArguments;
 }
 ```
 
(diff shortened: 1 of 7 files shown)


NoNameProvided:
Thanks for opening this PR, but this won't be merged as is. I see this is related to I18N which I want to add a well-rounded solution, not ad-hoc expanding the API.

Thanks for understanding.

PR #1025, opened 2021-04-15
feat: add IsStrongPassword decorator

### LINKED ISSUE #1024: feature: add IsStrongPassword decorator
### Description
Validating strength of passwords is a very common requirement for applications. At this moment `class-validator` doesn't support that even though `validator.js` already does.

### Proposed solution
Add the `IsStrongPassword` decorator to wrap the functionality already provided by `validator.js`.

jhonatanhulse:
## Description
Add support to password strength validation.

## Checklist
  - `npm run lint:check` passes

### Fixes
fixes #1024

### DIFF
--- a/src/decorator/string/IsStrongPassword.ts
+++ b/src/decorator/string/IsStrongPassword.ts
@@ -1,0 +1,42 @@
+import validator from 'validator';
+import { ValidationOptions } from '../ValidationOptions';
+import { buildMessage, ValidateBy } from '../common/ValidateBy';
+
+export const IS_STRONG_PASSWORD = 'isStrongPassword';
+
+/**
+ * Options to be passed to IsStrongPassword decorator.
+ */
+export type IsStrongPasswordOptions = Pick<
+  validator.StrongPasswordOptions,
+  'minLength' | 'minLowercase' | 'minUppercase' | 'minNumbers' | 'minSymbols'
+>;
+
+/**
+ * Checks if the string is a strong password.
+ * If given value is not a string, then it returns false.
+ */
+export function isStrongPassword(value: unknown, options?: IsStrongPasswordOptions): boolean {
+  return typeof value === 'string' && validator.isStrongPassword(value, options);
+}
+
+/**
+ * Checks if the string is a strong password.
+ * If given value is not a string, then it returns false.
+ */
+export function IsStrongPassword(
+  options?: IsStrongPasswordOptions,
+  validationOptions?: ValidationOptions
+): PropertyDecorator {
+  return ValidateBy(
+    {
+      name: IS_STRONG_PASSWORD,
+      constraints: [options],
+      validator: {
+        validate: (value, args): boolean => isStrongPassword(value, args.constraints[0]),
+        defaultMessage: buildMessage(eachPrefix => eachPrefix + '$property is not strong enough', validationOptions),
+      },
+    },
+    validationOptions
+  );
+}
--- a/README.md
+++ b/README.md
@@ -899,6 +899,7 @@ isBoolean(value);
 | `@IsISSN(options?: IsISSNOptions)`              | Checks if the string is a ISSN. |
 | `@IsISRC()`                                     | Checks if the string is a [ISRC](https://en.wikipedia.org/wiki/International_Standard_Recording_Code). |
 | `@IsRFC3339()`                                  | Checks if the string is a valid [RFC 3339](https://tools.ietf.org/html/rfc3339) date. |
+| `@IsStrongPassword(options?: IsStrongPasswordOptions)` | Checks if the string is a strong password. |
 | **Array validation decorators**                 | |
 | `@ArrayContains(values: any[])`                 | Checks if array contains all values from the given array of values. |
 | `@ArrayNotContains(values: any[])`              | Checks if array does not contain any of the given values. |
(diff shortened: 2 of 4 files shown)


rpCal:
@jhullse I just check code, its awesome! I just notice 
```ts 
validate: (value, args): boolean => isStrongPassword(value, args.constraints[0]) 
``` 

arg can be undefined and it can throw error... maybe you can update line to? 
```ts 
validate: (value, args): boolean => isStrongPassword(value, args?.constraints?.[0] ?? {})
```

rpCal:
@jhullse 
if you can please update also interface IsStrongPasswordOptions to contains optional params from validator.js isStrongPassword [LINK](https://github.com/validatorjs/validator.js)
```ts
export interface IsStrongPasswordOptions {
  ...(prev interface code)

  returnScore?: boolean;
  pointsPerUnique?: number;
  pointsPerRepeat?: number;
  pointsForContainingLower?: number;
  pointsForContainingUpper?: number;
  pointsForContainingNumber?: number;
  pointsForContainingSymbol?: number;
}
```

in the definition of isStrongPassword **If returnScore is true, then the function returns an integer score for the password rather than a boolean.** so return type should be boolean or number

```ts

export function isStrongPassword(
  value: unknown,
  options?: IsStrongPasswordOptions
): boolean | number {
  ...
}
```

jhullse:
Thanks for the feedback @rpCal. I'd like to clarify some things before I change the code.

---
**About the first suggestion:**
> arg can be undefined and it can throw error... maybe you can update line to?

When I was implementing the decorator I realized that `args` could be undefined looking at https://github.com/typestack/class-validator/blob/ff79d7dc27dd30b1f3f955b9948d1a8e4e8ab125/src/validation/ValidatorConstraintInterface.ts#L9

But at the same time I looked at several decorators in https://github.com/typestack/class-validator/tree/develop/src/decorator and they were all implemented assuming that `args` will be there and defined, I implemented this way to keep consistency. So I believe somehow the library is enforcing `args` to be available (I didn't find how exactly) and if that's the case maybe the `ValidatorConstraintInterface` needs to be changed. I'm OK with changing the `args` there but could you please give me an example of `@IsStrongPassword`'s usage where `args` will be undefined?

---
**About the second suggestion:**
> in the definition of isStrongPassword If returnScore is true, then the function returns an integer score for the password rather than a boolean. so return type should be boolean or number

When I was implementing the decorator the scoring part didn't seem to fit because if you use `@IsStrongPassword` you will never get back the score so why have an option in the decorator that says `returnScore` if the score won't be returned? Of course the user could use `isStrongPassword` function instead but we would be polluting `IsStrongPasswordOptions` interface, so maybe two different interfaces would make sense, one for the decorator (without scoring) and another for the function (with scoring). With two different interfaces we can avoid 

```
          if (typeof validationResult == "boolean") {
            return validationResult;
          } else {
            return validationResult > 0;
          }
```

which would cause the decorator to always return true even with weak passwords. What do you think?

dakotacookenmaster:
I am eagerly awaiting this! Let's get this added ASAP - I know other developers who would be very interested in having this built in to class-validator.

Jogiter:
nice work @jhullse. after a few months, this is still not merged. i have checked this pr:

**remove changes in package.json will fix the conflicts.** 

i really nead this feature. so to ask would u update it. great thanks to u.

jhonatanhulse:
@Jogiter done. The problem was that Dependabot updated some dependencies. The original PR didn't change any dependencies so I basically rebased my fork and now things look good.

src/decorator/string/IsStrongPassword.ts
```diff
@@ -0,0 +1,64 @@
...
+   */
+  minUppercase?: number;
+
+  /**
+   * The minimum amount of numbers.
+   */
+  minNumbers?: number;
+
+  /**
+   * The minimum amount of symbols.
```
braaar:
```suggestion
   * The minimum amount of special characters.
```

I think this is the more common terminology for special symbols like`#`, `%`, and so on.

However, I can understand the desire to keep the consistent naming convention since this is referred to as "symbols" in validator.js

Jogiter (dismissed):
LGTM

src/decorator/string/IsStrongPassword.ts
```diff
@@ -0,0 +1,64 @@
+import { ValidationOptions } from '../ValidationOptions';
+import { buildMessage, ValidateBy } from '../common/ValidateBy';
+import isStrongPasswordValidator from 'validator/lib/isStrongPassword';
+
+export const IS_STRONG_PASSWORD = 'isStrongPassword';
+
+/**
+ * Options to be passed to IsStrongPassword decorator.
+ */
+export interface IsStrongPasswordOptions {
```
NoNameProvided: You don't need to re-add the interfaces, you can directly import it from `valdator.js`. See other decorators for an example.

NoNameProvided (changes requested):
If you are still interested in getting this merged, then please rebase to the latest develop.

NoNameProvided:
Did you close by accident?

NoNameProvided (changes requested):
Thanks for the rebase! Tests and build are currently failing.

jhonatanhulse:
@NoNameProvided The version of`validator.js` was updated and I didn't notice, that is why the test failed. I believe it should be working now. I believe it needs approval in order to run the workflow again

nosachamos:
Hey guys, really looking forward to this one - anyone able to merge it? Seems like it's all green :)

jhonatanhulse:
@NoNameProvided I believe there is a change request from you that is already fixed (tests were failing before but not anymore) but it is still blocking the PR. When you have time, could you take a look at it? I'm not sure if @braaar or @Jogiter can help with this

NoNameProvided (approved):

PR #1048, opened 2021-05-01
fix: Nested validations optimization

### LINKED ISSUE #343: perf: nested validation is slow
# Problem
Validation of a plain object is really fast. 
Once an object is nested within another it gets "really" slow
- Simple single field validation: 0ms
- Nested 2 level single field validation: 50ms

# Question
Can we somehow make it faster? Seems like it doesn't have to be that much more complex?

# Example
```typescript
@Exclude()
class NestedChildValidator {
  @Expose()
  @IsString()
  public text: string;
}

@Exclude()
class ChildValidator {
  @Expose()
  @ValidateNested({each: true})
  @Type(() => NestedChildValidator)
  @ArrayMinSize(1)
  public children: NestedChildValidator[];
}

@Exclude()
class ParentValidator {
  @Expose()
  @ValidateNested({each: true})
  @Type(() => ChildValidator)
  @ArrayMinSize(1)
  public children: ChildValidator[];
}
```

```typescript
it('Should throw nested validation error', async () => {
  const val: ParentValidator = {
    children: [{ children: undefined }]
  };

  const input = plainToClass(ParentValidator, val);
  try {
    console.log(Date.now());
    await validate(input);
    expect(false, 'Validator should throw and not continue').to.equal(true);
  } catch (error) {
    console.log(Date.now()); // 50ms after first console log
  }
});

it('Should throw simple validation error', async () => {
  const val: NestedChildValidator = {
    text: undefined
  };

  const input = plainToClass(NestedChildValidator, val);
  try {
    console.log(Date.now());
    await validate(input);
    expect(false, 'Validator should throw and not continue').to.equal(true);
  } catch (error) {
    console.log(Date.now()); // Same time as first console log
  }
});
```

aaleksandrov:
## Description

This optimizes performance of all validations and closes https://github.com/typestack/class-validator/issues/343.

For example in my company we have 50+ complex models so `MetadatStorage.validationMetadatas` is array of ~5000 elements, `MetadatStorage.constraintMetadatas` is an array of 2500 elements. Validator loops over these arrays multiple times, so it takes 5+ seconds to validate a batch of 550 objects.

## Checklist

  - `npm run lint:check` passes

Didn't add any tests because my PR is pure refactoring of `MetadatStorage`, functionality is covered by tests already
### Fixes

fixes #343

---------------------

Running tests in my company's project normally
```
  V1BulkProductsSpec
    ✓ should ingest 550 products as bulk (5938ms)

  1 passing (6s)
```

Running tests in my company's project with this PR (42% performance improvement)
```
   V1BulkProductsSpec
    ✓ should ingest 550 products as bulk (3451ms)
    
  1 passing (3.5s)
```


dotslashx:
Just bumped into the same issue myself. I wonder why there is no pressure or event a slight interest for this case. The performance improvement is huge even for applications with decent amount of validation decorators. Lets say my application has 30 data classes, every class contains 10 fields and every field has 3 decorators. Then, there are these two simple classes which I'm having 2000 instances of and I'm trying to validate them.

My data classes:
```typescript
class Employee {
    @IsDefined()
    @IsInt()
    public id: number;

    public constructor() {
        this.id = 321;
    }
}

class Company {
    @IsDefined()
    @IsArray()
    @ArrayNotEmpty()
    @IsObject({each: true})
    @ValidateNested()
    public employees!: Employee[];

    public constructor() {
        this.employees = [...Array(10)].map(() => new Employee());
    }
}
```
My test script:
```typescript
console.info('Generating 2000 Company objects');
const companies = [...Array(2000)].map(() => new Company());

console.info('Validating 2000 Company objects');
let startDate = Date.now();

await Promise.all(companies.map(async item =>
    validate(item, {whitelist: true, forbidNonWhitelisted: true}),
));

let endDate = Date.now();
console.info(`Validation of 2000 Company objects took ${endDate - startDate} milliseconds`);

// THIS IS ONLY TO SIMPLIFY THIS EXAMPLE
console.info('Generating 30 dummy classes with 10 fields each, 3 decorators per field');
for (let dummyId = 0; dummyId < 30; dummyId++) {
    const dummy = class {};

    for (let propertyId = 0; propertyId < 10; propertyId++) {
        Object.defineProperty(dummy, `property_${propertyId}`, () => `property_${propertyId}`);

        for (let decoratorId = 0; decoratorId < 3; decoratorId++) {
            registerDecorator({target: dummy, name: `decorator_${decoratorId}`, propertyName: `property_${propertyId}`, validator: () => true});
        }
    }
}

console.info('Validating 2000 Company objects');
startDate = Date.now();

await Promise.all(companies.map(async item =>
    validate(item, {whitelist: true, forbidNonWhitelisted: true}),
));

endDate = Date.now();
console.info(`Validation of 2000 Company objects took ${endDate - startDate} milliseconds`);
```

The results I'm getting:
```
Generating 2000 Company objects
Validating 2000 Company objects
Validation of 2000 Company objects took 108 milliseconds
Generating 30 dummy classes with 10 fields each, 3 decorators per field
Validating 2000 Company objects
Validation of 2000 Company objects took 1486 milliseconds
```

And here are the results I'm getting with a patch from this PR applied:
```
Generating 2000 Company objects
Validating 2000 Company objects
Validation of 2000 Company objects took 98 milliseconds
Generating 30 dummy classes with 10 fields each, 3 decorators per field
Validating 2000 Company objects
Validation of 2000 Company objects took 139 milliseconds
```

This patch gave me 10 times performance improvement!

b26:
Bump

braaar:
@aaleksandrov, could you rebase/merge this and get rid of the irrelevant changes in this PR?

aaleksandrov:
Hi @braaar 
I merged `develop` into my branch and resolved conflicts. However as you can see meanwhile `develop` was force-pushed so my PR still contains random commits. Do you want to me fix this (maybe creating a new PR would be the simplest way) or it's ok since changes are correct?

braaar:
> Hi @braaar I merged `develop` into my branch and resolved conflicts. However as you can see meanwhile `develop` was force-pushed so my PR still contains random commits. Do you want to me fix this (maybe creating a new PR would be the simplest way) or it's ok since changes are correct?

In my opinion, the diff is all that matters and squashing and merging is the better way to deal with PRs, but I'm not sure if that is the opinion of the maintainers.

src/metadata/MetadataStorage.ts
```diff
@@ -35,14 +35,26 @@ export class MetadataStorage {
    * Adds a new validation metadata.
    */
   addValidationMetadata(metadata: ValidationMetadata): void {
-    this.validationMetadatas.push(metadata);
+    const existingMetadata = this.validationMetadatas.get(metadata.target);
+    if (existingMetadata) {
+      existingMetadata.push(metadata);
+      this.validationMetadatas.set(metadata.target, existingMetadata);
```
Clashsoft, line 41: This looks unnecessary, the key->value combination already exists as push does not change the array identity.

src/metadata/MetadataStorage.ts
```diff
@@ -35,14 +35,26 @@ export class MetadataStorage {
...
 
   /**
    * Adds a new constraint metadata.
    */
   addConstraintMetadata(metadata: ConstraintMetadata): void {
-    this.constraintMetadatas.push(metadata);
+    const existingMetadata = this.constraintMetadatas.get(metadata.target);
+    if (existingMetadata) {
+      existingMetadata.push(metadata);
+      this.constraintMetadatas.set(metadata.target, existingMetadata);
```
Clashsoft, line 54: Same here

src/metadata/MetadataStorage.ts
```diff
@@ -102,7 +115,13 @@ export class MetadataStorage {
     });
 
     // get metadatas for inherited classes
-    const inheritedMetadatas = this.validationMetadatas.filter(metadata => {
+    const filteredForInheritedMetadatasSearch = [];
+    for (const [key, value] of this.validationMetadatas.entries()) {
+      if (targetConstructor.prototype instanceof key) {
+        filteredForInheritedMetadatasSearch.push(...value);
+      }
+    }
```
Clashsoft, line 123: Isn't it faster to loop over the prototype chain and look up the contructors in the map?

NoNameProvided:
You have merged back the diverged branch in your PR, as a result, I cannot clean this, I have cherry-picked your commit and opened a new PR. 

Superset by https://github.com/typestack/class-validator/pull/1825

src/metadata/MetadataStorage.ts
NoNameProvided: This is fixed in the new PR.

src/metadata/MetadataStorage.ts
NoNameProvided: This is fixed in the new PR.

src/metadata/MetadataStorage.ts
NoNameProvided: I think you are right, but this I did not fix in the new PR. I just merged it in its current state for the performance gains. If you are up for the task, please implement it in a new PR.

Issue #1090, opened 2021-05-24
feature: Support unknown and known values in one object when whitelist and forbid turned on


AckerApple:
### Description
We have a `metadata` object that has both known values and unknown values. Our project has both `forbidUnknownValues: true` and `whitelist: true`. We couldn't find a way to allow unknown values in our `metadata` because the global validation kicks in. We couldn't just remove `@ValidateNested()` because we still wanted validation on some known properties.

### Proposed solution
I created a decorator to just `validate` without the inclusion of global rules

Creating PR now

PR #1091, opened 2021-05-24
Add decorator ValidateNoGlobals to validate without global configs involved

AckerApple:
## Description
Provides decorator to validate without the inclusion of global configurations. Greatly assist when a project has `forbidUnknownValues: true`, `whitelist: true`, and also has an object with both known and unknown values

## Checklist
  - `npm run lint:check` passes
- [-] tests are added for the changes I made (if any source code was modified)

### Fixes
Relates to #1090 

### Additional context

I'm not saying this is the fix we should be accepting. It works for us and I would like to put into hands that may perfect, enhance, or provide a more sound solution.

**Example of solution**:
```
import { IsEnum, IsNotEmpty, Validate } from 'class-validator'
import { Type } from 'class-transformer'

export class SomeStripeConnector {
  @Validate(NoGlobalValidate) // skip whitelist and unknown values
  @IsNotEmpty() @Type(() => MetaData)
  metadata: MetaData
}

export enum Env {
  DEV: 'dev'
  BETA: 'beta
  PRODUCTION: 'prod'
}

export class MetaData implements IMetaData {
  @IsEnum(Env)
  env: Env
}

export interface IMetaData {
  env: Env
  [index: string]: string
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -518,6 +518,40 @@ When skipping missing properties, sometimes you want not to skip all missing pro
 for you, even if skipMissingProperties is set to true. For such cases you should use `@IsDefined()` decorator.
 `@IsDefined()` is the only decorator that ignores `skipMissingProperties` option.
 
+## Including specific unknown values
+
+Read this section if your project has global protection via `forbidUnknownValues: true`, `whitelist: true`, and your project also has a need to have an object that accepts both known and unknown values. The technique used here is to invoke validation without the inclusion of global configurations.
+
+```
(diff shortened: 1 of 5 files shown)


NoNameProvided:
This won't get merged in its current form. 

I understand the reason you are adding this. In its current form `class-validator` is not suited to validate dynamic objects. However, I don't think this is the correct approach to the problem. 

We need further discussion and agreement on the solution before implementation.

PR #1097, opened 2021-05-26
fix: passing options when using @IsDateString()

### LINKED ISSUE #1061: fix: option is not passed to validator in `@IsDateString` decorator
### Description
According to the type docs, when using `IsDateString({ strict: true})`:
"If `strict` is `true`, performs additional checks for valid dates, e.g. invalidates dates like `2009-02-29`". But those dates appear to be passing the validation in my app

So I created a repo for bug reproduction [here](https://github.com/GCastilho/class-validator-strict-date-bug/tree/master). In essence:

```ts
class ValidatorTest {
	@IsDateString({ strict: true })
	isoDate: string
}

const instance = new ValidatorTest()
instance.isoDate = '2009-02-29'

await assert.rejects(() => validateOrReject(instance), 'Class successfully validated') // <<<< Fails
```

### Expected behavior
According to the docs, a date as "2009-02-29" should fail when using isoDate validation with strict = true; (taken directly from the docs)

### Actual behavior
The validator passes

### Version
class-validator: 0.13.1
node: 14

hakimio:
## Description

`@IsDateString({strict: true})` doesn't work because options are not passed to the validator.

### Fixes

Fixes #1061

### DIFF
--- a/src/decorator/string/IsDateString.ts
+++ b/src/decorator/string/IsDateString.ts
@@ -24,7 +24,7 @@ export function IsDateString(
       name: IS_DATE_STRING,
       constraints: [options],
       validator: {
-        validate: (value, args): boolean => isDateString(value),
+        validate: (value, args): boolean => isDateString(value, args.constraints[0]),
         defaultMessage: buildMessage(
           eachPrefix => eachPrefix + '$property must be a valid ISO 8601 date string',
           validationOptions
--- a/package-lock.json
+++ b/package-lock.json
@@ -1059,13 +1059,13 @@
       "dev": true
     },
     "@typescript-eslint/eslint-plugin": {
-      "version": "4.22.1",
-      "resolved": "https://registry.npmjs.org/@typescript-eslint/eslint-plugin/-/eslint-plugin-4.22.1.tgz",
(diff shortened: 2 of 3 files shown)


src/decorator/string/IsDateString.ts
```diff
@@ -24,7 +24,7 @@ export function IsDateString(
       name: IS_DATE_STRING,
       constraints: [options],
       validator: {
-        validate: (value, args): boolean => isDateString(value),
+        validate: (value, args): boolean => isDateString(value, args.constraints[0]),
```
braaar, line 27: This should work and reads better, no?

```suggestion
        validate: (value, args): boolean => isDateString(value, options),
```

braaar (changes requested):

braaar:
Would you care to rebase/merge this to get rid of merge conflicts and irrelevant changes?

GCastilho:
Any reason any it was closed?

braaar:
I would assume @hakimio is going to make a fresh PR instead of fixing this one. It's a bit cleaner that way, and it makes it easier to pass the [unnecessarily strict commit message check](https://github.com/typestack/class-validator/issues/1524)

hakimio:
PR #1720

PR #1125, opened 2021-06-10
Update "No metadata found" error to be more correct and to describe a second reason that metadata is not found

### LINKED ISSUE #1124: fix: Rewrite "No metadata found" error more clearly, adding explanation that perhaps a using codebase is implemented but no flags have been written
### Description
`src/validation/ValidationExecutor.ts` conditionally returns the following error: `No metadata found. There is more than once class-validator version installed probably. You need to flatten your dependencies.` There are two issues with this error.
- the error has typos/incorrect grammar. "There is more than once class-validator version installed probably" should instead read "There are probably more than one class-validator versions installed", or "There are probably multiple class-validator versions installed"
- This error sometimes arises when a developer uses a module as a dependency that itself depends on `class-validator`, but the developer has not yet used a `class-validator` flag in their file or codebase. One such case is documented in a [type-graphql issue](https://github.com/MichalLytek/type-graphql/issues/150), for example.

### Proposed Solution
Update the error message with correct grammar and add a sentence explaining that validation flags are expected but not present in the code.

capndave:
## Description

Update the following error: `No metadata found. There is more than once class-validator version installed probably. You need to flatten your dependencies.` 

- Update the error message with correct grammar and add a sentence explaining that validation flags are expected but not present in the code.

## Checklist

  - `npm run lint:check` passes

### Fixes
fixes #1124

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,8353 @@
 {
   "name": "class-validator",
   "version": "0.13.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.13.1",
+      "license": "MIT",
+      "dependencies": {
+        "@types/validator": "^13.1.4",
+        "libphonenumber-js": "^1.9.19",
+        "validator": "^13.6.0"
+      },
+      "devDependencies": {
+        "@rollup/plugin-commonjs": "^19.0.0",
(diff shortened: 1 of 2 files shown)


capndave:
Closing. Opening separate PR with semantic commit and pr title

capndave:
@jackjennings @bogdan @dooman87 @nolazybits how do I get this to pass the semantic PR? Have been fooling with it for over an hour and still won't pass

PR #1301, opened 2021-09-18
fix: add predicate for isBoolean

### LINKED ISSUE #1300: fix: add predicate for isBoolean
### Description
Hey, I have found one little bug in types. `isBoolean` should have a predicate which states, that `value is boolean`.

Here is a simple example. TypeScript will keep thinking, `val` is `123`, but `isBoolean` should make him think, its boolean.

```ts
function getRandomType(): string | number | boolean {
  return 'No random in our life';
}

const val = getRandomType(); // val here is string, or number, or boolean

if (isBoolean(val)) {
  const someAnotherVal = val; // someAnotherVal will have type string, or number, or boolean, but it should be clear boolean
}
```

### Expected behavior
TypeScript knows that value has type `boolean`.

### Actual behavior
`isBoolean` returns boolean value without predicate, so, it does not say anything for TypeScript.

heyqbnk:
## Description
Add predicate for `isBoolean` validator.

## Checklist

  - `npm run lint:check` passes

### Fixes

fixes #1300

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5094,9 +5094,9 @@
       "dev": true
     },
     "path-parse": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.6.tgz",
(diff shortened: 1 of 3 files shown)


braaar:
@heyqbnk could you rebase/merge this to get rid of irrelevant changes to package.json?

NoNameProvided:
This is already fixed in develop.

PR #1390, opened 2021-11-11
fix: #438 default forbidUnknownValues to true

### LINKED ISSUE #438: fix: default settings allows arbitrary bypass vulnerability
With this vulnerability, an attacker can bypass any security checks enforced by class-validator. 

When class-validator is used to validate user-input,  the attributes in the user-input object will be transformed into the validation class instance. 
However, the transforming procedure will overwrite the internal attribute of validation class instance (e.g., constructor attribute) if the attacker injects an attribute with the same name into user-input. Once this internal attribute being overwritten, class-validator will be bypassed.

PoC
```

import {validate, validateOrReject, Contains, IsInt, Length, IsEmail, IsFQDN, IsDate, Min, Max} from "class-validator";
import {plainToClass} from "class-transformer";

class Post {

    @Length(10, 20)
    title: string;

    @Contains("hello")
    text: string;

    @IsInt()
    @Min(0)
    @Max(10)
    rating: number;

    @IsEmail()
    email: string;

    @IsFQDN()
    site: string;

    @IsDate()
    createDate: Date;

}

let userJson = JSON.parse('{"title":1233, "__proto__":{}}');  // a malformed input
let users = plainToClass(Post, userJson);

validate(users).then(errors => { // errors is an array of validation errors
    if (errors.length > 0) {
        console.log("validation failed. errors: ", errors);
    } else {
        console.log("validation succeed");
    }
});

```

Our suggestion is that class-validator should check the integrity of the constructor: if it is being corrupted, the validation should automatically fail.

mikeguta:
## Description
This is an attempted fix for https://github.com/typestack/class-validator/issues/438
It does so with a breaking change that uses a safe default of `true` for `forbidUnknownValues`. Only an explicit `false` value would preserve the previous default behaviour.

## Checklist
  - `npm run lint:check` passes

### Fixes
fixes #438 

### Other
2f3f241ba52e065cd8bdc735c7ebeb3b3a6f5cde Moved `@types/validator` to `devDependencies`
6b5d1d567654d60da21cbdade8dc425b5c64f7fd Fixed a couple of tests that were not returning promises and failing silently

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ export interface ValidatorOptions {
 }
 ```
 
-> It's highly advised to set `forbidUnknownValues: true` as it will prevent unknown objects from passing validation.
+> It's highly advised to not set `forbidUnknownValues: false` as it will allow unknown objects to bypass validation. The default is a safe value of `true`.
 
 ## Validation errors
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1154,7 +1154,8 @@
     "@types/validator": {
       "version": "13.6.6",
       "resolved": "https://registry.npmjs.org/@types/validator/-/validator-13.6.6.tgz",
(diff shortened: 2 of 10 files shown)


mikeguta:
will replace with with another PR

PR #1400, opened 2021-11-16
Prevent silently failing tests by waiting results

Dimrok:
## Description

Await checks (`checkInvalidValues`, `checkValidValues`) to get the actual result from jest and not false positive and dangling errors.

### Before
image

### After
image

## Checklist

  - `npm run lint:check` passes

### Fixes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5094,9 +5094,9 @@
       "dev": true
     },
     "path-parse": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.6.tgz",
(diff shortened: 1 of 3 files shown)


NoNameProvided:
Closing this as I think this changes nothing. Returning a promise or creating an async function and awaiting the inside is the same thing from the consumer's point of view. Eg:

```ts
it('test case', () => {
  return checkInvalidValues(new MyClass(), invalidValues, { skipUndefinedProperties: true });
});
// Equal to the following  
it('test case', async () => {
  await checkInvalidValues(new MyClass(), invalidValues, { skipUndefinedProperties: true });
});
```

The screenshots display failed assertions which were caused by using incorrect internal names in some of the tests, but that is fixed already.

PR #1502, opened 2022-01-20
docs(IsDecimal): use correct docstring for IsDecimal

### LINKED ISSUE #1423: Does IsDecimal do what it should do?
Why `IsDecimal()` decorator description says that "Checks if the string contains only letters and numbers. If given value is not a string, then it returns false." and it does not check of the value was decimal or not?

While `isDecimal()` contains proper description and probably does that it means.

For example `@IsDecimal` on a field with value 123.45 passes validation while it should not, imo.

```ts
import { ValidationOptions } from '../ValidationOptions';
import ValidatorJS from 'validator';
export declare const IS_DECIMAL = "isDecimal";
/**
 * Checks if the string is a valid decimal.
 */
export declare function isDecimal(value: unknown, options?: ValidatorJS.IsDecimalOptions): boolean;
/**
 * Checks if the string contains only letters and numbers.
 */
export declare function IsDecimal(options?: ValidatorJS.IsDecimalOptions, validationOptions?: ValidationOptions): PropertyDecorator;
```

braaar:
## Description
Use the correct docstring for the decorator `IsDecimal`. The existing one seems to have been copied from `IsAlphanumeric`:
[IsAlphanumeric source code](https://github.com/typestack/class-validator/blob/develop/src/decorator/string/IsAlphanumeric.ts)

## Checklist

  - `npm run lint:check` passes
- [ ] I have run the project locally and verified that there are no errors

### Fixes
fixes #1423

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "eslint-plugin-jest": "^25.3.3",
         "husky": "^4.3.8",
         "jest": "^26.6.3",
-        "lint-staged": "^12.1.4",
+        "lint-staged": "^12.1.5",
         "prettier": "^2.5.1",
         "reflect-metadata": "0.1.13",
         "rimraf": "3.0.2",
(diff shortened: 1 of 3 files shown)


braaar:
I think the merge commit made the Semantic Pull Request check fail?

braaar:
As long as we are using squash and merge I don't see any issue with the PR.

NoNameProvided:
Landed in https://github.com/typestack/class-validator/commit/42b9ca263cf855d2004c0dce23e38d23546148ac

PR #1697, opened 2022-07-15
feat: Support for external message composing

eeemzs:
Created a new Validator option field, externalMessageComposer, to support returning the supplied validation parameters inside the decorator back as an error message to allow users to create their standard custom messages, and allow them to easily create localized error messages. It is no longer the only option to write custom error messages every time using a decorator which breaks the DRY principle; the new feature helps clean code for custom error messages.

## Description

Sample Code:

```
export class DtoTest {
  @MinLength(30)
  title: string;
}

const dto = new DtoTest();
dto.title = "Length of this messsage is 29";

const options = {
  externalMessageComposer: false,
};

validate(dto, options).then((errors: string | any[]) => {
  // errors is an array of validation errors
  if (errors.length > 0) {
    console.log("validation failed. errors: ", errors);
  } else {
    console.log("validation succeed");
  }
});

```
Before the feature support: Resulting Error Message inside ValidationError is:

```
    constraints: {
      minLength: 'title must be longer than or equal to 30 characters'
    }
```

Now if we set externalMessageComposer to true:

```
const options = {
  externalMessageComposer: true,
};

```
After the feature support and the externalMessageComposer is set to true: 
Resulting Error Message inside ValidationError is now:

`constraints: { minLength: '30' }`

By this way; we can compose our message, as it was previosly not possible to get the validation parameter ('30').

For example; user may compose the following error message as the {1} argument ('30' in this example) is supplied within the class-validator error message.

` "minLength": "{0} must be longer than or equal to {1} characters",`

## Checklist

  - `npm run lint:check` passes

### Fixes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "eslint-plugin-jest": "^25.3.3",
         "husky": "^4.3.8",
         "jest": "^26.6.3",
-        "lint-staged": "^12.1.4",
+        "lint-staged": "^12.1.5",
         "prettier": "^2.5.1",
         "reflect-metadata": "0.1.13",
         "rimraf": "3.0.2",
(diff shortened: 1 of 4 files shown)


NoNameProvided (changes requested):
I don't think we want to have a property with a mixed return type. Instead, we can return this data always under a new property. 

I see this PR has no related discussion, before opening a PR please always open a discussion, detailed feature request, or bug report. PRs without related issue may be closed automatically in the future.

NoNameProvided:
To clarify, I think we should implement returning the constraint metadata under a new key, but before you open a new PR please open a discussion first.

eeemzs:
Done! Thank you for the suggestion.
Discussion [#1800](https://github.com/typestack/class-validator/discussions/1800)

Issue #1719, opened 2022-08-16
Is this project in need of more maintainers?


braaar:
As far as I can tell, the [last pull request to get merged](https://github.com/typestack/class-validator/pull/830) was one and a half years ago. Since then, the project has seen a few bug fixes from @NoNameProvided and a lot of automated bumps, but there have been several high quality PRs opened and reviewed by the community. 

Last year, several people [raised concerns](https://github.com/typestack/class-validator/issues/178#issuecomment-831178151) over the level activity here, and I believe there are many highly qualified and dedicated open source developers who would be willing to take on more responsibility on this project. 

I'm relatively fresh in the game, so I don't know how this process usually works, but my thinking is that I want to make it as easy as possible for the current maintainers to find good candidates. I think the best way to show that you are dedicated and willing to take such a responsibility is by contributing wherever you can, even with the boring stuff. 

I would encourage people who are frustrated with the lack of activity here to go through unresolved issues and try to resolve them and get their authors to close them. We have a backlog to get through! (353 issues at the time of writing, and uhh, now it's 354 with this one)

I ❤️ `class-validator`

braaar:
I suppose it's relevant to tag the other Typestack folks: @saulotoledo @jotamorais @MichalLytek @rustamwin

eeemzs:
I agree with you. I have had a pull request for a good to have feature as well, then after it is approved, I need to apply this feature support on NestJS repository, going somewhere with two tickets.  But apperantly this will take months! It is too sad for such a beatiful project is not actively maintained.

tciuro:
Re: "and get their authors to close them"

That's the key. Looks like the authors are no longer interested? There hasn't been any feedback from @saulotoledo @jotamorais @MichalLytek or @rustamwin since you asked on Aug. 16, now over a month ago. Total radio silence from the team. 🤔

I was looking for an issue I saw with `$value` not passing through in the validation and saw [this PR](https://github.com/typestack/class-validator/pull/1467) waiting to be approved since... Dec. 27, 2021. Wow. It actually seems to fix two issues: https://github.com/typestack/class-validator/issues/1046 and https://github.com/typestack/class-validator/issues/921

braaar:
> and get their authors to close them

By this I meant getting people to close their own issues since we have no active maintainers to do it for them.

tciuro:
Ah! Got it.

Do you know if it's being discussed to fork the project and continue development?

jotamorais:
[...]

PR #1720, opened 2022-08-17
fix: pass options to validator in `@IsDateString` decorator

### LINKED ISSUE #1061: fix: option is not passed to validator in `@IsDateString` decorator

hakimio:
## Description

Fixes calling `@IsDateString({strict: true})`. Currently, it doesn't work because the options are not passed to the underlying validator.

### Fixes

references #1061

### DIFF
--- a/src/decorator/string/IsDateString.ts
+++ b/src/decorator/string/IsDateString.ts
@@ -24,7 +24,7 @@ export function IsDateString(
       name: IS_DATE_STRING,
       constraints: [options],
       validator: {
-        validate: (value, args): boolean => isDateString(value),
+        validate: (value): boolean => isDateString(value, options),
         defaultMessage: buildMessage(
           eachPrefix => eachPrefix + '$property must be a valid ISO 8601 date string',
           validationOptions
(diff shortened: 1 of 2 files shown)


braaar (dismissed):

hakimio:
@braaar you don't have permission to merge?

braaar:
> @braaar you don't have permission to merge?

Nope, I'm just a humble member of the community :)

I'm trying to get this repo active again, though. See #1719

hakimio:
At this point you might just fork the project. It doesn't seem like @NoNameProvided has any plans to work on this anymore.

braaar:
I'm hesitant to take it that far. With the amount of stars and weekly downloads this project has, it would be a huge setback and inconvenience for the end users. That's a last resort if the entire Typestack team goes completely AWOL.

NoNameProvided (approved):

PR #1773, opened 2022-10-26
fix stopAtFirstError not working bug

Ives7:
Set the stopAtFirstError property to true, and the rest of the validation will still be performed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "eslint-plugin-jest": "^25.3.3",
         "husky": "^4.3.8",
         "jest": "^26.6.3",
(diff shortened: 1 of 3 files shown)


ytetsuro:
@Ives7 

Please, write test code.

Current stopAtFirstError option has [test code](https://github.com/typestack/class-validator/blob/6c501131d537c0ca608f651708a5c87cf1b5312a/test/functional/validation-options.spec.ts#L1210-L1214).
It's all green.

Please let we know following.

* Happen code
* Expected
* Actual

NoNameProvided:
Closing due to no reply from the PR author.

PR #1817, opened 2022-11-29
Change object type in `ValidationArguments` to Record<any, any>

### LINKED ISSUE #1816: fix: Accessing keys inside `object` property that's in `ValidationArguments` Interface
### Description

When trying to access a specific key to an object that is in `ValidationArguments` interface, I'm getting the following error
```
TS7053: Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{}'.
```
which is make sense because the object property inside `ValidationArguments` takes an object as a type and the [object type](https://www.typescriptlang.org/docs/handbook/basic-types.html#object) is meant to abstract away any keys of an object,
```
	async validate(value: string, argumentes: ValidationArguments) {
		const [relatedPropertyName] = argumentes.constraints;
		const relatedValue = argumentes.object[relatedPropertyName]; // typescript complaining 
	}
```

### Expected behavior
Typescript shouldn't complain :) 

### Actual behavior

Typescript error
`TS7053: Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{}'.`

AbdlrahmanSaberAbdo:
## Description

Replaced `object` type to `Record<any, any>` to object property inside ValidationArguments interface, whereas [Record<K, T>](https://github.com/Microsoft/TypeScript/blob/af8e44ac85a46ead640b3204a6e4c333641cf945/lib/lib.es5.d.ts#L1379-L1384) exists to specifically define the keys of a type. This means there is a difference when trying to access object properties.

TypeScript will allow to access any property of an object of type Record<any, any> even though the specific keys are not known, since the first generic parameter is any.

We can replace the first key with `string` if we want to be more strict.

On an object of type object however, the keys are not assumed to be any. As with Record<any, ...>, TypeScript does not know which keys actually exist, but it will not allow to access any keys:
```
let b: object;
a.foo; // error: Property "foo" does not exist on type "object"
```

## Checklist

  - `npm run lint:check` passes

### Fixes

fixes #1816


NoNameProvided:
This won't be merged as is. The validated object is not a `Record` type but anything. If we would change this to anything it would be `unknown` type. 

I understand you want to access a property on it. You should cast it to your preferred type when accessing it in your custom decorator.

NoNameProvided:
Closing discussion here, to redirect it to the issue. Please share your reply there.

PR #1825, opened 2022-12-03
perf: store metadata in maps instead of arrays

### LINKED ISSUE #343: perf: nested validation is slow

NoNameProvided:
## Description

From the original PR:
> This optimizes the performance of all validations and closes https://github.com/typestack/class-validator/issues/343.
>
> For example in my company we have 50+ complex models so `MetadatStorage.validationMetadatas` is array of ~5000 elements, `MetadatStorage.constraintMetadatas` is an array of 2500 elements. The validator loops over these arrays multiple times, so it takes 5+ seconds to validate a batch of 550 objects.

## Checklist

  - `npm run lint:check` passes
- [ ] I have run the project locally and verified that there are no errors

### Fixes

references https://github.com/typestack/class-validator/issues/343

### DIFF
--- a/src/metadata/MetadataStorage.ts
+++ b/src/metadata/MetadataStorage.ts
@@ -12,11 +12,11 @@ export class MetadataStorage {
   // Private properties
   // -------------------------------------------------------------------------
 
-  private validationMetadatas: ValidationMetadata[] = [];
-  private constraintMetadatas: ConstraintMetadata[] = [];
+  private validationMetadatas: Map<any, ValidationMetadata[]> = new Map();
+  private constraintMetadatas: Map<any, ConstraintMetadata[]> = new Map();
 
   get hasValidationMetaData(): boolean {
-    return !!this.validationMetadatas.length;
+    return !!this.validationMetadatas.size;
   }
 
   // -------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Clashsoft (commented):
See https://github.com/typestack/class-validator/pull/1048#pullrequestreview-1202857343

PR #1966, opened 2023-02-27
Add improved validation message for unknown property

Recodify:
## Description
Gives you a clue as to what property in the source object caused `an unknown value was passed to the validate function`

Before: `an unknown value was passed to the validate function`
After: `an unknown value was passed to the validate function for property user`


tsa96:
This looks like a useful change, but do you realize you've included a 10k lines of package-lock.json changes? I don't know what the review process on this repo looks like but I'd recommend just `reset --mixed`ing your branch and recommitting without all that, then force-pushing.

PR #1972, opened 2023-02-28
feat: new Validation decorators IsNatural and IsNaturalNoZero

yasinpalizban:
Hi i add two new functionality to decorators called  IsNatural and IsNaturalNoZero.

here is simple definition for them 

IsNatural  if field contains anything other than a natural number: 0, 1, 2, 3, etc.  

iSNaturalNoZero if field contains anything other than a natural number, except zero: 1, 2, 3, etc.


Clashsoft:
Why when you can just combine `@IsInt` and `@IsPositive`?

PR #2099, opened 2023-05-20
Remove functions from enum values

SimonRosenau:
## Description
This change removes function declarations from enum values.

I am using an enum with an companion namespace with utility functions:
```typescript
export enum ResponseTypeEnum {
  Code = 'code',
}

export namespace ResponseTypeEnum {
  export const fromString = (value: string): ResponseTypeEnum => {
    switch (value.toLowerCase()) {
      case 'code':
        return ResponseTypeEnum.Code
      default:
        throw Error('Invalid response type: ' + value)
    }
  }
}
```

Within a request body as such:
```typescript
export class Request {
  @IsEnum(ResponseTypeEnum)
  responseType: ResponseTypeEnum
}
```

The error returned by the server upon sending an invalid enum value is serialized as the following:
```
"responseType must be one of the following values: code, (value) => {\n        switch (value.toLowerCase()) {\n            case 'code':\n                return ResponseTypeEnum.Code;\n            default:\n                throw Error('Invalid response type: ' + value);\n        }\n    }"
```

This PR removes the function declaration from the possible enum values in the error message

## Checklist

  - `npm run lint:check` passes

### DIFF
--- a/src/decorator/typechecker/IsEnum.ts
+++ b/src/decorator/typechecker/IsEnum.ts
@@ -17,6 +17,7 @@ export function isEnum(value: unknown, entity: any): boolean {
 function validEnumValues(entity: any): string[] {
   return Object.entries(entity)
     .filter(([key, value]) => isNaN(parseInt(key)))
+    .filter(([key, value]) => typeof value !== 'function')
     .map(([key, value]) => value as string);
 }
 
--- a/src/decorator/typechecker/IsEnum.ts
+++ b/src/decorator/typechecker/IsEnum.ts
@@ -15,9 +15,7 @@ export function isEnum(value: unknown, entity: any): boolean {
  * Returns the possible values from an enum (both simple number indexed and string indexed enums).
  */
 function validEnumValues(entity: any): string[] {
-  return Object.entries(entity)
-    .filter(([key, value]) => typeof value === 'string' && isNaN(parseInt(key)))
-    .map(([key, value]) => value as string);
+  return Object.values(entity).filter(value => typeof value === 'string') as string[];
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Clashsoft:
Why not combine the filter operations to save an intermediate array creation?

SimonRosenau:
@Clashsoft In this case I would prefer code style in favor of performance, as honestly the performance gain would be quite minuscule, but if others would agree that this should be refactored I would also be open to do so

Clashsoft:
In addition I would even argue that the parseInt should be guarded with typeof value === 'string' instead of !== function because that is the assumption of `as string` in the last line. So some change is necessary anyway

SimonRosenau:
Hey @Clashsoft! Sorry for the long wait. I have addressed your requested changes, but now I'm not even sure whether the `isNaN(parseInt(key))` check is even needed anymore?

Clashsoft:
@SimonRosenau Thanks for addressing! I believe you don't need it anymore. See https://stackoverflow.com/a/43091709/4138801

> ```ts
> Object.keys(myEnum).map(key => myEnum[key]).filter(value => typeof value === 'string') as string[];
> ```

Or with ES2017 according to [this comment](https://stackoverflow.com/questions/18111657/how-to-get-names-of-enum-entries#comment85271383_43091709):

> ```ts
> Object.values(myEnum).filter(value => typeof value === 'string') as string[];
> ```

SimonRosenau:
@Clashsoft Makes absolute sense! Adjusted 🙂

Clashsoft:
LGTM!

SimonRosenau:
@Clashsoft GHA needs your approval it seems 🙂

Clashsoft:
I can't, Im not a maintainer

Issue #1992, opened 2023-03-13
feature: validate Emoji


svyatoslavratov:
Example:
```js
class ExampleDto {
  @IsEmoji()
  public emoji: string;
}
```
Expected work like this:
```js
const reg = /\p{Emoji}/u;

reg.test("😀"); // true
reg.test("U+1F600"); // true
```

PR #2130, opened 2023-06-15
feat: Add IsEmoji validator

edilson:
## Description

This PR aims to add a new validator to check if the character passed is actually an emoji. It attends the issue #1992.

This will only validate if the passed is the emoji itself not the unicode character:

```TS
class ExampleDto {
  @IsEmoji()
  public emoji: string;
}

```

If I have such a class and I pass to it this value: ```🫵🏾``` it should return true but if I pass the unicode ```1FAF5 1F3FE``` it throws an error because it should be an actual emoji and the regular expressions we have for such purpose don't validate the hex characters, only the emoji.

## Checklist
  - `npm run lint:check` passes


Clashsoft:
Bad PR. You forgot a console.log, the isEmoji function does unsafe casts before checking if its a string, it requires a new transitive dependency, and the whole thing could just be done with `@Matches(emojiRegex())`.

PR #2166, opened 2023-07-16
Patch IsNumber decorator with maxDecimalPlaces option set for numbers in scientific annotation

### LINKED ISSUE #1705: fix: IsNumber decorator fails on values ABS(x) < 0.000001 when using maxDecimalPlaces
Hi team,

`IsNumber` decorator throws error, when a small value is analyzed (ABS(x) < 0.000001. It happens because such a small value has exponential format -->

```bash
> parseFloat('0.000001')
< 0.000001
> parseFloat('0.0000001')
< 1e-7
```

The following code fails, because `value.toString().split('.')[1]` is not available - there's no `.` in the string!
```
  if (options.maxDecimalPlaces !== undefined) {
    let decimalPlaces = 0;
    if (value % 1 !== 0) {
      decimalPlaces = value.toString().split('.')[1].length;
    }
    if (decimalPlaces > options.maxDecimalPlaces) {
      return false;
    }
  }
```

https://github.com/typestack/class-validator/blob/2ef8ff0f32be53e75dade2a0b86b158d87427176/src/decorator/typechecker/IsNumber.ts#L34

backnight:
## Description

The **IsNumber()** decorator doesn't work on scientifically annotated numbers when specifying `maxDecimalPlaces` option because `value.toString()` has no dot in it and crashes (see issue: https://github.com/typestack/class-validator/issues/1705).

https://github.com/typestack/class-validator/blob/4a0ec6680161fc611ba9fdea74fd34653c3997e6/src/decorator/typechecker/IsNumber.ts#L34

This patch reworks the whole handling of `maxDecimalPlaces` option in order to make it work with scientific numbers.

The logic behind the patch is `toFixed(maxDecimalPlaces)` the input so it is represented as a string with a maximum number of decimals as specified by the option `maxDecimalPlaces`, then `parseFloat()` it to be compared with the original input.
If there is a difference with the original input, it means it has a number of decimals > `maxDecimalPlaces` because they have been truncated by `toFixed(maxDecimalPlaces)`

## Checklist

- [ ] the code follows the established code style of the repository
  - `npm run lint:check` passes

### Fixes

fixes #1705

### DIFF
--- a/src/decorator/typechecker/IsNumber.ts
+++ b/src/decorator/typechecker/IsNumber.ts
@@ -29,11 +29,7 @@ export function isNumber(value: unknown, options: IsNumberOptions = {}): value i
   }
 
   if (options.maxDecimalPlaces !== undefined) {
-    let decimalPlaces = 0;
-    if (value % 1 !== 0) {
-      decimalPlaces = value.toFixed(options.maxDecimalPlaces).split('.')[1].length;
-    }
-    if (decimalPlaces > options.maxDecimalPlaces) {
+    if (parseFloat(value.toFixed(options.maxDecimalPlaces)) !== value) {
       return false;
     }
   }
--- a/src/decorator/typechecker/IsNumber.ts
+++ b/src/decorator/typechecker/IsNumber.ts
@@ -29,8 +29,10 @@ export function isNumber(value: unknown, options: IsNumberOptions = {}): value i
   }
 
   if (options.maxDecimalPlaces !== undefined) {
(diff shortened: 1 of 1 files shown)


skoniks (commented):
`parseFloat(value.toFixed(...)) !== value` works as expected
Waiting for approvement ✔

maierthomas:
also waiting... 😒

backnight:
ChatGPT Image Jun 11, 2025, 11_21_07 PM

PR #2189, opened 2023-08-03
Improve code for IsOptional validation

Branchverse:
Improved null/undefined checking code

## Description
No functional change, just a better and faster and most importantly cool solution so our code can be fancier, faster and shorter.
Don't be a pleb, be a double caster.

same logic: 
image

This performs around 6 times faster than the previous code: [benchmark](https://jsbench.me/ztlkvfqos0/1)
image

## Checklist
  - `npm run lint:check` passes


Clashsoft:
This changes the semantics. Suddently 0, false, empty string all cause other validators to be skipped.

Issue #1721, opened 2022-08-17
feature: Conditional validation inside validation options


soroushmadani:
### Description
Sometimes we need to validate a field based on another field in the dto.

### Proposed solution
1. it would be nice to have an option called `validateIf` which is a callback function with given object.
example:

 ``` js
export class TestDto {
  // ...

  @IsNotEmpty()
  @IsString()
  type: string;

  @IsUUID({ validateIf: o => o.type === "something" })
  @IsEmpty({ validateIf: o => o.type === "something" })
  @IsNotEmpty({ validateIf: o => o.type !== "something" })
  albumId: string;
}
```

2. the above example would do the job for me, but it would be nicer if we could group validation somehow like below:

 ```js
export class TestDto {
 // ...

  @ValidationGroup([IsUUID(), IsEmpty()], { validateIf: o => o.type === "something" })
  @IsNotEmpty({ validateIf: o => o.type !== "something" })
  albumId: string;
}
```
or:
 ```js
export class TestDto {
 // ...

  @ValidationGroup(o => {
    if (o.type === "something" ) {
      return [IsUUID(), IsEmpty()]
    }
    return [IsNotEmpty()]
  })
  albumId: string;
}
```

In general, it's really good to access current object when validating. for example I needed to set the max value of a number based on another field. I made my own decorator which works like:

```js
 @Max((o) =>
    o.type === DiscountPlanType.Percentage ||
      ? 100
      : false, // when false is passed, this will be skipped
  )
```

braaar:
One way to achieve what you want today is to use the existing [validation groups](https://github.com/validatorjs/validator.js/pull/2009) combined with a function that, before validating, checks some property of the object to decide which groups to validate. Once you have figured that out, you can then call `validate` with the groups passed in the `ValidatorOptions`.

I'm not saying this to discard your suggestion, but this is a basis for comparison. What are the strengths and weaknesses of these approaches compared to one another? Personally I don't use groups, so I don't have a strong opinion about them. 

I would suspect that with your approach it's easy for someone reading your code (either someone else or you in the future) to struggle to understand the intricacies of the logic since it would be scattered about in various decorators. Especially your last suggestion. The flexibility of this feature means that it's hard to make sure that you are programming with a consistent pattern. I am the type of person who would be reluctant to add features that, while powerful, could encourage bad practices and messy code.

4ybakut2004:
That would be a very useful feature.

[...]

PR #1579, opened 2022-03-11
feat: `validateIf` for validation options

aoi-umi:
## Description

Add `validateIf` for validation options.
If you want to validate that condition by object, you can use `validation validateIf`.

`ValidateIf` will skip all other decorators, `validation validateIf` only effect the decorator that use this option

## Checklist

- [x] the pull request title describes what this PR does (not a vague title like `Update index.md`)
- [x] the pull request targets the *default* branch of the repository (`develop`)
- [x] the code follows the established code style of the repository
  - `npm run prettier:check` passes
  - `npm run lint:check` passes
- [x] tests are added for the changes I made (if any source code was modified)
- [x] documentation added or updated
- [x] I have run the project locally and verified that there are no errors

## Fixes
references https://github.com/typestack/class-validator/issues/1455

PR #2244, opened 2023-09-16
feat: add option to validate decorators separately

### LINKED ISSUE #318: Why Conditional Validation ignored all constraint on a field instead?
I am trying to create multi purpose model using conditional validation. Look at the below model -

```
class TestModel extends Model {

    public isCreated: boolean | undefined;
    @IsUUID()
    @ValidateIf(o => o.isCreated === true)
    @IsNotEmpty()
    public id: string | undefined;
}
```

i want that if isCreated is false then id must behave as optional parameter. and IsUUID() constraint should not be ignored. ValidateIf ignore both constraint. Any Idea?

mario-orlicky:
## Description
Added a new validation option which forces to always validate a property even if the conditional validation results false. There is already an option named `always` which works with groups so I decided to name the new one `ignoreValidation`.

All tests pass except 2 tests related to JWT, but those tests fail also on develop branch so it was not caused by my changes - this could be caused by upgrading package versions regularly.

## Checklist
  - `npm run lint:check` passes

### Fixes
fixes #318


tareksalem:
I don't know why a PR like this is not reviewed until now, I don't see the library maintainers willing to develop the library or add anything new, this library is so limited and good only for simple projects!!

braaar:
See [my recent comment](https://github.com/typestack/class-validator/issues/1721#issuecomment-1857290706) in #1721

braaar:
implemented in https://github.com/typestack/class-validator/pull/1579

PR #2360, opened 2024-01-10
fix: validator type imports for latest @types/validator

### LINKED ISSUE #1866: Does IsMobilePhone accept an array ?
**I was trying to...**
`
 @IsMobilePhone(['sk-SK', 'sr-RS'])
  mobile: string;
`
**The problem:**

it show a message which is 'type string[] can’t  pass to type MobilePhoneLocale'. but in isMobilePhone.d.ts, the annotation says 'locale is either an array of locales, or one of  ...'

`
 * Checks if the string is a mobile phone number (locale is either an array of locales (e.g ['sk-SK', 'sr-RS'])
 * OR one of ['am-Am', 'ar-AE', 'ar-BH', 'ar-DZ', 'ar-EG', 'ar-IQ', ar-JO', 'ar-KW', 'ar-SA', 'ar-SY', 'ar-TN', 'be-BY', ...
`

so how can i use IsMobilePhone and pass an array to param locale

### LINKED ISSUE #68152: [@types/validator] Cannot find namespace 'ValidatorJS'. Did you mean 'validator'?
13.11.8 introduced build errors. I had to use 13.11.7 instead of 13.11.8.
`Error: ../../node_modules/class-validator/types/decorator/string/IsAlpha.d.ts(8,58): error TS2833: Cannot find namespace 'ValidatorJS'. Did you mean 'validator'?`

_Originally posted by @kabamrrichman in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/68121#issuecomment-1884113872_

benasher44:
## Description
In https://github.com/DefinitelyTyped/DefinitelyTyped/pull/68121, we fixed validator's types to improve compatibility with node16/nodenext. This fixes missing information that was important to users under certain configurations, but it unfortunately means that type imports in the repo need to be updated to be namespace imports (i.e. can no longer rely on the global validator namespace). This workaround is the way forward.

## Checklist

  - `npm run lint:check` passes

### Fixes

Closes DefinitelyTyped/DefinitelyTyped#68152
fixes #1866

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.14.0",
       "license": "MIT",
       "dependencies": {
-        "@types/validator": "^13.9.0",
+        "@types/validator": "^13.11.8",
         "libphonenumber-js": "^1.10.53",
         "validator": "^13.9.0"
       },
@@ -1716,9 +1716,9 @@
       "dev": true
     },
     "node_modules/@types/validator": {
-      "version": "13.9.0",
-      "resolved": "https://registry.npmjs.org/@types/validator/-/validator-13.9.0.tgz",
-      "integrity": "sha512-NclP0IbzHj/4tJZKFqKh8E7kZdgss+MCUYV9G+TLltFfDA4lFgE4PKPpDIyS2FlcdANIfSx273emkupvChigbw=="
+      "version": "13.11.8",
+      "resolved": "https://registry.npmjs.org/@types/validator/-/validator-13.11.8.tgz",
+      "integrity": "sha512-c/hzNDBh7eRF+KbCf+OoZxKbnkpaK/cKp9iLQWqB7muXtM+MtL9SUUH8vCFcLn6dH1Qm05jiexK0ofWY7TfOhQ=="
     },
     "node_modules/@types/yargs": {
       "version": "17.0.13",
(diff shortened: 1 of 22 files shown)


benasher44:
@braaar from looking at the GitHub issues, it looks like this style of namespace import issue [comes up every few years](https://github.com/typestack/class-validator/issues?q=is%3Aissue+namespace+ValidatorJS). This should fix it permanently for this repo.

braaar:
This is great! I've been meaning to take a look at what's going on with the types that cause issues like #1866.

braaar:
I notice you are not bumping the validator package itself. Isn't there a correspondence between versions of `@types/valitor` and `validator` that we should respect?

braaar (approved):
Since the checks are green, I see no reason to hold back on this. Thanks!

benasher44:
> I notice you are not bumping the validator package itself. Isn't there a correspondence between versions of `@types/valitor` and `validator` that we should respect?

Thanks for merging! Generally, yes. As it stands, this library doesn't use any new functions introduced since the declared version of validator. If you decide to introduce code that uses newer validator functionality, you'll need to bump the validator version to say that "hey if you're not using at least this newer version, it might crash". Since no new runtime functionality was introduced, I left the validator version alone, but I'm happy to open a new PR to bump it if you'd like!

For types, technically this library only needs the types from before this PR, since it only uses APIs/types from at least that version. However, this PR introduces a new goal wherein I want to show that when you update to at least this types version, there is an error, so then this PR can prove that error existed and then is fixed. To showcase that, I bumped the types version.

The new types version is backward compatible for the existing APIs you're already using in validator package, but you'll need to update the validator package version should you decide to broadcast usage of any of the newer APIs allowed by the added types in the newer types package version (though again happy to open a PR to preemptively upgrade).

I hope this makes sense!

Otherwise, might you be able to publish a new release? I think the folks on the linked ticket would appreciate it!

benasher44:
Saw the release — thank you!

PR #2443, opened 2024-03-27
feat: add IsNullable decorator

### LINKED ISSUE #491: @IsOptional with @IsNotEmpty is not working

Motii1:
## Description
Added `IsNullable` decorator

## Checklist

  - `npm run lint:check` passes

### Fixes

partially fixes [#491 ](https://github.com/typestack/class-validator/issues/491)


braaar:
I would try solving this with @ValidateIf or https://github.com/typestack/class-validator/pull/1579

Issue #2577, opened 2025-01-27
feature: standardschema support


niemyjski:
It would be awesome to support this and go a long ways towards a better dx story. https://standardschema.dev/

braaar:
I doubt this will be making it to class-validator. I think there is a fundamental incompatibility between class decorator based validation and schema based validation. Schema based validation was intentionally dropped from the project, not because schema based validation is a bad idea, but because there are other projects that do that (I use zod myself in all my projects and love it).

If there is a great way to map class decorators to schemas and I'm just not seeing, let me know! Not having used NestJS or class based validator for several years may cloud my perspective on this.

niemyjski:
I think it's possible but lots of work. I love this library but I've moved to zod as I couldn't use this with tanstack forms and many other libraries due to missing standard schema.

van-sprundel:
+1, Zod has become the de-facto for a lot of projects and if this is not on the roadmap then I would argue it's going to pass up on a real community voice.

braaar:
Well, alright. If someone wants to do it, I won't stand in the way.

PR #2588, opened 2025-03-10
feat: add standard schema support and associated tests

GregoireBellon:
## Description
This PR adds the support for standard schema, by copying their interface, extending the validator with it, and mapping class validator's "Validation Error" type to standard schema's "Issue". 

I also wrote some tests, which are the same as the "nested validation" ones.

Since this function would be in the hot path, I tried to make it optimized, by avoiding recursion and preferring mutation to copy.

Supporting the Standard Schema interface is important, since it's blocking the adoption of it by NestJs (see related [nestjs/nest/#14539](https://github.com/nestjs/nest/issues/14539))

edit: by itself, the PR in its current form would not be enough. We need to add a wrapper function, that takes a schema as input, and that outputs a validator containing the schema. Let me know if it sounds good to you.

## Disclosure
I had to ignore the ESLint warning when copying the standard schema interface.

## Checklist
  - `npm run lint:check` passes

### Fixes
addresses #2577

### DIFF
--- a/src/standard-schema/StandardSchema.ts
+++ b/src/standard-schema/StandardSchema.ts
@@ -1,0 +1,110 @@
+/**
+ * The Standard Schema interface.
+ */
+export type StandardSchemaV1<Input = unknown, Output = Input> = {
+  /**
+   * The Standard Schema properties.
+   */
+  readonly '~standard': StandardSchemaV1.Props<Input, Output>;
+};
+
+export declare namespace StandardSchemaV1 {
+  /**
+   * The Standard Schema properties interface.
+   */
+  export interface Props<Input = unknown, Output = Input> {
+    /**
+     * The version number of the standard.
+     */
+    readonly version: 1;
+    /**
+     * The vendor name of the schema library.
+     */
+    readonly vendor: string;
+    /**
+     * Validates unknown input values.
+     */
+    readonly validate: (value: unknown) => Result<Output> | Promise<Result<Output>>;
+    /**
+     * Inferred types associated with the schema.
+     */
+    readonly types?: Types<Input, Output> | undefined;
+  }
+
+  /**
+   * The result interface of the validate function.
+   */
+  export type Result<Output> = SuccessResult<Output> | FailureResult;
+
+  /**
+   * The result interface if validation succeeds.
+   */
+  export interface SuccessResult<Output> {
+    /**
+     * The typed output value.
+     */
+    readonly value: Output;
+    /**
+     * The non-existent issues.
+     */
+    readonly issues?: undefined;
+  }
+
+  /**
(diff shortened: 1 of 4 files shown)


0x0bit:
@braaar @NoNameProvided

src/standard-schema/StandardSchema.ts
```diff
@@ -0,0 +1,110 @@
...
+   */
+  export interface SuccessResult<Output> {
+    /**
+     * The typed output value.
+     */
+    readonly value: Output;
+    /**
+     * The non-existent issues.
+     */
+    readonly issues?: undefined;
```
niemyjski, line 50: Is this type correct?

niemyjski:
@braaar @NoNameProvided This would be massive if this can be supported. See discussion here https://github.com/nestjs/nest/issues/14539 for possible rework?

niemyjski:
@braaar @NoNameProvided Any updates would be greatly appreciated and or appointing some new maintainers.

niemyjski:
What's left on this? Can we get this merged?

src/standard-schema/StandardSchema.ts
GregoireBellon: This is the standard schema interface, [as the official doc](https://standardschema.dev/) specifies it

GregoireBellon:
Hello @niemyjski, sorry for not responding earlier, I kinda lost interest in this PR since I don't think it will ever be merged. 

This is not fully ready. Making class validator compliant will require adding a new API. 

Citing my comment on the [Nest compatibility with standard schema](https://github.com/nestjs/nest/issues/14539#issuecomment-2717079063) issue :

>Standard Schema assumes that the validator is closely tied to the schema it uses to validate a DTO, and that a DTO can be a standard JavaScript object.
>
> Class-validator's validator stores all validation constraints in a singleton context called "ValidationMetadata" and it relies on the DTO's constructor to identify the validation constraints related to the DTO. This means a DTO must be an instance of a class.
> 
> This difference in philosophy means that class-validator will need additional APIs to support Standard Schema's approach.

Since nobody is currently governing the project, I don't believe that any significant change in the API will be approved.

niemyjski:
Thanks for your reply, I really like this library at this point I think I might have to fork or move to zod :(

braaar:
I doubt this will be making it to class-validator. I think there is a fundamental incompatibility between class decorator based validation and schema based validation. Schema based validation was intentionally dropped from the project, not because schema based validation is a bad idea, but because there are other projects that do that (I use zod myself in all my projects and love it).

braaar:
I was a little harsh earier. This seems like it could be good for the project and would be greatly appreciated by those who like this library.

Issue #686, opened 2020-08-10
how to exclude null from @IsOptional


horrendo:
I'm trying to validate a field which can be missing (or undefined), but if present, cannot be null. Currently it seems that if I have the `@IsOptional()` decorator and the value is null, it bypasses any further validation.

I tried making my own decorator similar to `@IsOptional()` but got a runtime error:

```
Error: Cannot find module 'class-validator/types/metadata/ValidationMetadata'
Require stack:
{my modules}
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:965:15)
    at Function.Module._resolveFilename (/Users/stbaldwin/git/unsullied/node_modules/tsconfig-paths/lib/register.js:75:40)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Object.<anonymous> (/Users/stbaldwin/git/unsullied/src/utils/validation/OptionalButNotNull.ts:3:1)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Module.m._compile (/Users/stbaldwin/git/unsullied/node_modules/ts-node/src/index.ts:837:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/stbaldwin/git/unsullied/node_modules/ts-node/src/index.ts:840:12)
```

Here is my decorator:

```typescript
import { ValidationOptions, ValidationTypes, getMetadataStorage } from 'class-validator';
import { ValidationMetadataArgs } from 'class-validator/types/metadata/ValidationMetadataArgs';
import { ValidationMetadata } from 'class-validator/types/metadata/ValidationMetadata';
/**
 * Checks if value is missing and if so, ignores all validators.
 */
export function IsOptionalButNotNull(validationOptions?: ValidationOptions): PropertyDecorator {
    return function (object: object, propertyName: string | symbol): void {
        const args: ValidationMetadataArgs = {
            type: ValidationTypes.CONDITIONAL_VALIDATION,
            target: object.constructor,
            propertyName: String(propertyName),
            constraints: [
                (object: any, _value: any): boolean => {
                    return object[propertyName] !== undefined;
                }
            ],
            validationOptions: validationOptions
        };
        getMetadataStorage().addValidationMetadata(new ValidationMetadata(args));
    };
}
```

[...]

PR #2632, opened 2025-09-18
feat: add `@IsStrictlyOptional()` decorator filtering only missing `undefined` values

ehenon:
## Description
As mentioned/discussed in some issues and PRs, proposal to add a new decorator based on the same philosophy as `@IsOptional()`, but filtering only values that are actually missing (`=== undefined`). Equivalent to `@ValidateIf((o, v) => v !== undefined)`, but more direct and explicit.

## Checklist
  - `npm run lint:check` passes

## References
references [#491](https://github.com/typestack/class-validator/issues/491), [#2615](https://github.com/typestack/class-validator/pull/2615), [#686](https://github.com/typestack/class-validator/issues/686)


AntonioJRM1998:
I think it makes more sense to use this decorator in the way I suggested. My view of “optional” in a DTO is: if the property is not present in the request body, validation should be skipped; if it is present (even if null or undefined), validation should run.

Using propertyName in object achieves this behavior:

Property absent → skip validation

Property present → run validation

This way, the decorator better reflects the semantic meaning of “optional” and avoids the confusion of skipping validation only for undefined, which doesn’t fully cover real-world use cases.

yaquawa:
yes, I need this too.

PR #2647, opened 2025-11-30
feat: update UUID validation to support versions 1-8, nil, max, and all

### LINKED ISSUE #2581: feature: 🔍 Add support for UUID v7 in `IsUUID` validator
### **Description**  
Currently, the `IsUUID` validator in `class-validator` only supports **UUID versions 3, 4, and 5**, but does not support **UUID v7**.  

UUID v7 is gaining adoption due to its **chronological ordering and improved indexing performance**, making it a preferred choice for databases and APIs. Many modern applications are switching to UUID v7, and its absence in `class-validator` makes validation inconsistent with newer standards.

---

### **Proposed solution**  
It would be great if `class-validator` could **add support for UUID v7** in the `IsUUID` validator.  

Since UUID v7 is already supported by `validator.js`, it should be possible to extend the existing validation to include it.

---

### **Additional context**  
- `validator.js` (which `class-validator` relies on) **already supports UUID v7**, so it would be great to make use of this.  
- UUID v7 is **designed for better performance and scalability**, especially in databases that require efficient indexing.  
- Adding UUID v7 support would ensure `class-validator` remains compatible with modern UUID standards.  

Would love to know if this is planned!  
Thanks for all the great work on `class-validator` 🚀

### LINKED ISSUE #2648: fix: isUUID validation fails for UUID like strings
### Description

Recently a breaking change was introduced in validator [v13.15.0](https://github.com/validatorjs/validator.js/releases/tag/13.15.0) which made UUID validation strict by also validating the variant field. The recent update of class-validator brought this breaking change of validator library in 0.14.3 causing failures for UUID like ID validations.

https://github.com/validatorjs/validator.js/pull/2421

```ts
isUUID('240d8c21-a21e-4282-0674-176658866702') // false since 0.14.3
```

### Expected behavior
UUID Validation for UUID like strings should pass when in `loose` mode, introduced in [validator@13.15.15](https://github.com/validatorjs/validator.js/releases/tag/13.15.15). `loose` mode should be properly exposed for consumption.

```ts
isUUID('240d8c21-a21e-4282-0674-176658866702', 'loose') // true
```

### Actual behavior

There is no `loose` mode to opt-into resulting in failed validations.

PleBea:
## Description

Updated the `IsUUID` tests to align with the updated JSDoc, which specifies support for all UUID versions (1-8, nil, max, all).

Previously, the tests primarily covered versions 3, 4, and 5. I have added test cases to verify that the `IsUUID` decorator correctly validates all UUID specifications supported by the underlying `validator.js` library, including versions 1, 2, 6, 7, 8, nil, and max.

## Checklist

  - `npm run lint:check` passes

### Fixes

fixes #2581 
fixed #2648


WikiRik:
Maybe #2648 can also be fixed with this PR? Since it's the same validator that requires an update

PleBea:
> Maybe #2648 can also be fixed with this PR? Since it's the same validator that requires an update

Yes, that's correct. This PR now specifically addresses issue #2648 as well.

I've made a small adjustment/tweak to the relevant code (the validator) to ensure it correctly fixes #2648. Once this PR is merged, we can close that issue.

test/functional/validation-functions-and-decorators.spec.ts
```diff
@@ -3719,6 +3726,387 @@ describe('IsUUID v5', () => {
...
+    '',
+    'xxxA987FBC9-4BED-3078-CF07-9141BA07C9F3',
+    '934859',
+    'AAAAAAAA-1111-1111-AAAG-111111111111',
+    'A987FBC9-4BED-3078-8F07-9141BA07C9F3', // v3
+    '713ae7e3-cb32-45f9-adcb-7c4fa86b90c1', // v4
+  ];
+
+  class MyClass {
+    @IsUUID('nil')
```
Setitch, line 3949: Question: In which situation i would set constant validation to nil uuid value? If i do that - its better to omit the field at all and put nil value directly further in the code instead.

test/functional/validation-functions-and-decorators.spec.ts
```diff
@@ -3719,6 +3726,387 @@ describe('IsUUID v5', () => {
...
+    '',
+    'xxxA987FBC9-4BED-3078-CF07-9141BA07C9F3',
+    '934859',
+    'AAAAAAAA-1111-1111-AAAG-111111111111',
+    'A987FBC9-4BED-3078-8F07-9141BA07C9F3', // v3
+    '713ae7e3-cb32-45f9-adcb-7c4fa86b90c1', // v4
+  ];
+
+  class MyClass {
+    @IsUUID('max')
```
Setitch, line 3990: Question: In which situation i would set constant validation to max uuid value? If i do that - its better to omit the field at all and put max value directly further in the code instead.

test/functional/validation-functions-and-decorators.spec.ts
PleBea: Good question! Here are some real-world use cases for nil UUID validation:

1. Databases without nullable UUID columns: Some databases (e.g., certain configurations of PostgreSQL, or when using ORMs with non-nullable UUID primary keys) use nil UUID as a sentinel value to represent "not set" instead of NULL.

2. UUID-based foreign key placeholders: When a foreign key relationship is optional but the column must be a valid UUID, nil UUID serves as the "no reference" value.

3. Distributed systems / Message queues: Some systems use nil UUID as a default/placeholder value in message schemas where the field is required but not yet populated.

4. API contract validation: When receiving data from external services that use nil UUID in their spec (e.g., some Microsoft APIs use nil GUID as a special value).

5. Completeness with validator.js: Since the underlying validator.js library already supports nil version, exposing it in class-validator maintains feature parity and gives users the full flexibility the library offers.

The key difference from omitting the field: omitting means "field doesn't exist" while nil UUID means "field exists with an explicit 'empty' value" - these have different semantic meanings in many systems.

test/functional/validation-functions-and-decorators.spec.ts
PleBea: You're right that max UUID has fewer practical use cases compared to nil. Here's my reasoning for including it:

1. RFC 9562 compliance: Both nil and max UUIDs are defined in RFC 9562 as special UUID values. Including both maintains compliance with the specification.

2. Feature parity with validator.js: The underlying validator.js library supports max as a version option. This PR aims to expose all the UUID validation capabilities that validator.js provides, rather than cherry-picking which ones to support.

3. Range queries: In some database scenarios, max UUID can be used as an upper bound for range queries on UUID columns (similar to how nil serves as lower bound).

4. Consistency: If we support nil (all zeros), it makes sense to also support max (all f's) for symmetry.

That said, I agree max is rarely used in practice. If you prefer, I can remove max (and potentially nil) from this PR and focus only on loose mode, which has clearer practical value. Let me know your preference!

test/functional/validation-functions-and-decorators.spec.ts
Setitch: Valid point, though why allowing only NIL value to be used? I mean, why not each? I think it would be better to have something like:  V7 + NIL + MAX,  V1 + NIL, V4 + MAX, instead of just MAX or just NIL.

PleBea: Good point! I'll add support for version arrays so you can validate against combinations like ['7', 'nil', 'max'] or ['1', 'nil']. The validation will pass if the UUID matches any of the specified versions.

braaar (approved):

PR #2651, opened 2025-12-05
Feat/add is username validator

Moses-Bassey:
## Checklist

- [ x] Added username validators that supports just some special strings
- [ x] Also enable users to define thier username string rules


Clashsoft:
Is this really that useful? It can be done with a regex validator already, and is highly subjective.

Setitch:
> Is this really that useful? It can be done with a regex validator already, and is highly subjective.

Totally agree, not only this adds not needed feature it also strips local language letters unless you provide them  - it would be just a pain in multi language environment and real life examples. 
One more note: user name is usually email address (especially with 3rd party login system).

braaar:
This is a bit specific to your use case. Use a [custom validator](https://github.com/typestack/class-validator?tab=readme-ov-file#custom-validation-decorators) instead

PR #2655, opened 2025-12-24
feat: add an option to IsObject to exclude functions from object checks

### LINKED ISSUE #2515: fix: isObject returns true for functions but docs say it should be false
### Description

lib version used:
`"0.14.1"`

jest test case
```ts
import { IsObject, isObject, validate } from 'class-validator';

it('BUG: isObject validates functions', () => {
  const fn = () => 'i am a function';

  expect(typeof fn).toBe('function');
  expect(isObject(fn)).toBe(false); // <<< this fails
});

class TestClass {
  @IsObject()
  public prop: unknown;
}

it('BUG: IsObject validates functions', async () => {
  const fn = () => 'i am a function';
  const subject = new TestClass();
  subject.prop = fn;

  expect(typeof subject.prop).toBe('function');
  const errors = await validate(subject);
  expect(errors).toEqual([]); // <<< this should fail but doesnt
});

```

Docs for `@IsObject()` read
> Checks if the object is valid Object (null, functions, arrays will return false).

I assume 

### Expected behavior

I personally would expect `isObject(fn)` to return `false`.

### Actual behavior

```log
$ jest
 FAIL  src/class-validator-bug.spec.ts
  ✕ BUG: isObject validates functions (3 ms)
  ✓ BUG: IsObject validates functions (2 ms)

  ● BUG: isObject validates functions

    expect(received).toBe(expected) // Object.is equality

    Expected: false
    Received: true

       5 |
       6 |   expect(typeof fn).toBe('function');
    >  7 |   expect(isObject(fn)).toBe(false); // <<< this fails
         |                        ^
       8 | });
       9 |
      10 | class TestClass {

      at Object.<anonymous> (class-validator-bug.spec.ts:7:24)

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 passed, 2 total
Snapshots:   0 total
Time:        1.508 s, estimated 2 s
Ran all test suites.
error Command failed with exit code 1.
```

## Investigation

https://github.com/typestack/class-validator/blob/63fe9c532fb6f633ddbb90e8225c5af49750513a/src/decorator/typechecker/IsObject.ts#L11 is the culprit.

niskyB:
## Description
IsObject returned true for functions despite the documented behavior. This PR resolves the issue by introducing an option to exclude functions from object checks.

## Checklist
  - `npm run lint:check` passes

### Fixes
fixes #2515


Setitch:
```
let a = () => {};
console.log(typeof a); // outcome => "function"
console.log(a instanceof Object); // outcome => true
```
Functions are objects in js therefore isObject should treat functions as objects by JS standard. 
However i presume additional option passed to decorator which would allow some control (exclude functions from being treated by the validator as objects) would be much better than going against the language principles.

niskyB:
Could you please check this again? @Setitch

braaar:
I don't think this is worth a breaking change. I would rather suggest you use https://github.com/typestack/class-validator/pull/1579 to cover your use case

PR #2661, opened 2026-01-13
release: 0.14.4

viplezer-well:
## Summary

- Updated validator.js to 13.15.22 ([#2649](https://github.com/typestack/class-validator/pull/2649))

## Changes

- docs: add changelog for 0.14.4
- build: bump version to 0.14.4

PR #2667, opened 2026-02-25
chore: release 0.14.4

braaar:
fixes #2661


braaar:
Thanks @viplezer-well!

